PR #28, opened 2025-09-15
feat: restructure infra components to streamline

unique-dominik:
Prepares for production deployments of the services out of experimental.

A connector comes with
```
service
├── deploy
│   ├── helm-charts
│   │   └── service
│   └── terraform-azure
│       ├── service-scope-module1
│       └── service-scope-module2
```

## helm-charts
is plural, you never know

## modules
differ per service, thus always are `service-name-XXXX` where XXXX is first a scope then a function.

## entra module
is not yet deployed and won't be until 5min before deployment where permissions and everything else are crisp and fresh and documented.

### DIFF
--- a/services/sharepoint-connector/deploy/terraform-azure/sharepoint-connector-entra-application-registration/variables.tf
+++ b/services/sharepoint-connector/deploy/terraform-azure/sharepoint-connector-entra-application-registration/variables.tf
@@ -1,6 +1,7 @@
 variable "display_name" {
   description = "The display name for the Azure AD application registration."
   type        = string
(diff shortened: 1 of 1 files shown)


unique-jakub (approved):
In general looks good, for the msgraph maybe we can start doing like this:
https://github.com/Unique-AG/hello-azure/blob/release/governance/dev/main.tf#L65-L73
https://github.com/Unique-AG/hello-azure/blob/release/terraform-modules/identities/application-registration.tf#L25

then we don't have to deal with the IDs

unique-dominik:
@unique-jakub we can, but legit question: how do you know its application or delegated?

PR #38, opened 2025-10-07
build(sharepoint-connector): Add Dockerfile

Ostrzy:


services/sharepoint-connector/deploy/Dockerfile
```diff
@@ -0,0 +1,67 @@
+# syntax=docker/dockerfile:1.7
+
+FROM node:22-bookworm AS base
+ENV PNPM_HOME="/pnpm"
```
lorand93, line 4: The connectors Dockerfile pins PNPM to version 10.15.1. We should do the same unless we have a reason

services/sharepoint-connector/deploy/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+COPY ./tsconfig.json ./tsconfig.json
+COPY ./turbo.json ./turbo.json
+COPY --from=pruner /app/out/full/ .
+
+RUN pnpm --filter=@unique-ag/${APP_NAME} codegen
+RUN --mount=type=cache,target=/root/.cache/turbo \
+    pnpm turbo build --filter=@unique-ag/${APP_NAME}... && \
+    pnpm --filter=@unique-ag/${APP_NAME} deploy out
+
+FROM node:22-bookworm-slim AS runner
```
lorand93, line 38: here as well, the existing dockerfile pins the pnpm version: ARG PNPM_VERSION=10.15.1

services/sharepoint-connector/deploy/Dockerfile
lorand93:
```
ARG PNPM_VERSION=10.15.1
RUN corepack enable && corepack prepare pnpm@${PNPM_VERSION} --activate
```

services/sharepoint-connector/deploy/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+FROM base AS builder
+ARG APP_NAME
+WORKDIR /app
+
+COPY --from=deps /app/ .
+COPY ./tsconfig.json ./tsconfig.json
+COPY ./turbo.json ./turbo.json
+COPY --from=pruner /app/out/full/ .
+
+RUN pnpm --filter=@unique-ag/${APP_NAME} codegen
```
lorand93, line 33: Just a remark 
Although this might be required when we add codegen, at the moment we don't really have a codegen at this step so this codegen cmd doesn't exist yet. 
Also the exissting dockerfile has this codegen step so we should be fine keeping it here.

services/sharepoint-connector/deploy/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+    ca-certificates \
+    curl \
+    && rm -rf /var/lib/apt/lists/* \
+    && groupadd --system --gid 1001 nodejs \
+    && useradd --system --uid 1001 --gid nodejs --shell /bin/bash --create-home nestjs
+
+ENV PNPM_HOME="/tmp/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+ENV COREPACK_HOME="/tmp/.cache/corepack"
+RUN corepack enable
```
lorand93, line 54: intentionally left out "RUN corepack prepare pnpm@${PNPM_VERSION} --activate" ?

services/sharepoint-connector/deploy/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+RUN mkdir -p /tmp/.cache/corepack && chown -R nestjs:nodejs /tmp/.cache/corepack
+
+ENV NODE_ENV=production
+
+COPY --from=builder --chown=nestjs:nodejs /app/services/${APP_NAME}/dist ./dist
```
lorand93, line 59: if we keep the codegen step we should also add ```COPY --from=builder --chown=nestjs:nodejs /app/services/${APP_NAME}/@generated ./@generated```

lorand93 (commented):
Looks great!

services/sharepoint-connector/deploy/Dockerfile
Ostrzy: I talked with @vabatta And he said that it's not required because `corepack` is doing it anyway based on `package.json` 🙂

services/sharepoint-connector/deploy/Dockerfile
Ostrzy: Yes, that's what @vabatta wanted to change 🙂

PR #57, opened 2025-10-08
fix(sharepoint-connector): fix non-string envs and avoid error swallowing

Ostrzy:


services/sharepoint-connector/deploy/helm-charts/sharepoint-connector/templates/config.yaml
```diff
@@ -9,22 +9,23 @@ data:
   ALLOWED_MIME_TYPES: {{ .Values.connectorConfig.allowedMimeTypes }}
   GRAPH_API_URL: {{ .Values.connectorConfig.graph.apiUrl }}
   GRAPH_TENANT_ID: {{ .Values.connectorConfig.graph.tenantId }}
-  GRAPH_USE_OIDC_AUTH: {{ .Values.connectorConfig.graph.useOidcAuth }}
-  PROCESSING_CONCURRENCY: {{ .Values.connectorConfig.processingConcurrency }}
+  GRAPH_USE_OIDC_AUTH: {{ .Values.connectorConfig.graph.useOidcAuth | quote }}
+  PROCESSING_CONCURRENCY: {{ .Values.connectorConfig.processingConcurrency | quote }}
```
Ostrzy, line 13: @unique-dominik So the configs we set up in values are anyway enforced as strings, BUT should we leave the quotes so that we are sure nothing wild from overrides breaks this? 🤔 Or whatever overrides are coming, we also validate them at some point? 🤔

unique-dominik: The GitHub Action prints the chart, we can check if theyre quoted
https://github.com/Unique-AG/connectors/actions/runs/18341437938/job/52237728119

services/sharepoint-connector/deploy/helm-charts/sharepoint-connector/values.schema.json
```diff
@@ -41,11 +41,11 @@
           "description": "Comma-separated list of allowed MIME types for file processing"
         },
         "processingConcurrency": {
-          "type": "integer",
+          "type": "string",
```
unique-dominik, line 44: A bit ugly but also okay for now. We should try to just quote just quick to learn something. I do locally.

unique-dominik (approved):

services/sharepoint-connector/deploy/helm-charts/sharepoint-connector/values.schema.json
unique-dominik: image
It works with just quote. Maybe we can leave the typing? WDYT?

services/sharepoint-connector/src/sharepoint-synchronization/sharepoint-synchronization.service.ts
```diff
@@ -44,11 +45,12 @@ export class SharepointSynchronizationService {
         );
 
         await this.orchestrator.processFilesForSite(siteId, files, diffResult);
-      } catch (error) {
-        this.logger.error(
-          `Failed during processing of site ${siteId}:`,
-          error instanceof Error ? error.stack : String(error),
-        );
+      } catch (rawError) {
+        const error = normalizeError(rawError);
```
lorand93, line 49: this should work

lorand93 (approved):

PR #332, opened 2026-03-06
fix(outlook-semantic-mcp,unique-api): remove hacky way to pass permissions

nicolae536:


packages/unique-api/src/clients/unique-graphql.client.ts
```diff
@@ -43,28 +43,18 @@ export class UniqueGraphqlClient {
             ...options,
             dispatcher: this.dispatcher,
           } as Parameters<typeof undiciFetch>[1],
-        )) as typeof fetch,
```
Ostrzy, line 46: Would be nice to comment why we need such hard type forcing 🙂

Ostrzy (approved):

packages/unique-api/src/clients/unique-graphql.client.ts
nicolae536: Done

PR #370, opened 2026-03-16
feat(confluence-connector,unique-api): implement content cleanup for deleted tenants

lorand93:
## Summary

- Implement content cleanup when a tenant's status is set to `deleted`. Deletes all ingested files and child scopes while preserving the root scope.
- Add `status` field to `NamedTenantConfig` and `TenantContext`. Config loader returns all tenants (active + deleted) in a single list.
- Every tenant (active or deleted) gets registered in a single map and gets its own cron job. The scheduler checks `tenant.status` at sync time to branch between cleanup and sync.
- Unified file deletion strategy: all tenants (V1 and V2) use scope-based content deletion via `getContentIdsByScope()` and `deleteByIds()`
- Add `getContentIdsByScope()` to unique-api files service for scope-based content queries
- Skip cleanup if root scope is not found or already empty (no child scopes)
- Startup assertion relaxed: connector can start with only deleted tenants (no active tenants required)

> **Simplification from previous version**: Removed the V1/V2 key format branching. Previously, V2 tenants used `deleteByKeyPrefix(tenantName)` while V1 tenants iterated child scopes. Now all tenants use the same scope-based deletion path, since all content is owned by child scopes regardless of key format. This eliminates the `useV1KeyFormat` dependency from `TenantDeleteService` and makes the constructor take a simple `scopeId` string instead of the full `IngestionConfig`.

**The judgement is that when and if we add support for Recursive ingestion we will add back the deleteByKeyPrefix method as by then we expect all clients to be on V2 Key format.**

## Architecture Diagram

```mermaid
flowchart TD
    A[App Startup] --> B[TenantRegistry.onModuleInit]
    B --> C[For each tenant config]
    C --> D{status?}
    D -->|active| E[Register full service stack]
    D -->|deleted| F[Register UniqueApiClient +<br/>TenantDeleteService only]
    E --> G[TenantSyncScheduler.onModuleInit]
    F --> G
    G --> H[Register cron job per tenant]
    H --> I[Cron tick calls syncTenant]
    I --> J{tenant.status?}
    J -->|active| K[ConfluenceSynchronizationService.synchronize]
    J -->|deleted| L[TenantDeleteService.deleteTenantContent]
```

## File Deletion Strategy

Uses a single unified approach for all tenants regardless of key format (V1 or V2):

```mermaid
flowchart TD
    A[Delete tenant content] --> B[List child scopes of root scope]
    B --> C{Has child scopes?}
    C -->|No| D[Skip - already cleaned]
    C -->|Yes| E[For each child scope]
    E --> F["getContentIdsByScope(scopeId)<br/>Query by ownerId"]
    F --> G{Has content?}
    G -->|Yes| H["deleteByIds(contentIds)"]
    G -->|No| I[Skip file deletion]
    H --> J[Next scope]
    I --> J
    J --> E
    J --> K[Delete all child scopes<br/>recursive: true]
    K --> L[Root scope preserved empty]
```

## Changed Files

| File | Change |
|------|--------|
| `services/confluence-connector/src/config/tenant-config-loader.ts` | Add `status` to `NamedTenantConfig`, return all tenants from `getTenantConfigs()`, remove `getDeletedTenantConfigs()` |
| `services/confluence-connector/src/config/index.ts` | Remove `getDeletedTenantConfigs` export |
| `services/confluence-connector/src/tenant/tenant-context.interface.ts` | Add `status: 'active' \| 'deleted'` field |
| `services/confluence-connector/src/tenant/tenant-registry.ts` | Single tenant map, branch on status for service registration, extracted helper methods |
| `services/confluence-connector/src/tenant/tenant-delete.service.ts` | Unified scope-based content deletion with `isScanning` re-entrancy guard |
| `services/confluence-connector/src/scheduler/tenant-sync.scheduler.ts` | Schedule all tenants, check status in `syncTenant()` to branch between cleanup and sync |
| `packages/unique-api/src/files/files.service.ts` | Add `getContentIdsByScope()` |
| `packages/unique-api/src/files/files.queries.ts` | Add `ContentByScopeInput` type |
| `packages/unique-api/src/files/unique-files.facade.ts` | Add method to interface |

## Test plan

- [x] `tenant-config-loader.spec.ts` - unified `getTenantConfigs()` returns active + deleted with status
- [x] `tenant-registry.spec.ts` - single map, status on contexts, deleted tenant service registration
- [x] `tenant-delete.service.spec.ts` - scope-based deletion, re-entrancy guard, partial failures
- [x] `tenant-sync.scheduler.spec.ts` - cron jobs for all tenants, status branching, error isolation
- [x] All 278 tests pass, lint and type check clean


Ostrzy:
Some legit points below 🙂 

## 🔍 Review Findings                                                                                                                                                         
                                                                                                                                                                              
  ---                                                                                                                                                                           
                                                                                                                                                                                
  ### ⚠️  Minor                                                                                                                                                                  
                                                                                                                                                                              
  - **Cleanup counters not initialized for deleted tenants**                                                                                                                    
    `tenant-registry.ts:100-113`
    `initializeCounters()` is only called for active tenants (line 184). The cleanup metrics (`cfc_cleanup_content_deleted_total`, `cfc_cleanup_scopes_deleted_total`) have no  
  `0` baseline, so Prometheus `increase()` alerts will produce `NaN` on the first scrape — the exact problem `initializeCounters` was designed to prevent.                      
   
  - **"Already cleaned up" path records a success duration metric**                                                                                                             
    `tenant-delete.service.ts:43-56`                                                                                                                                          
    The early return when `childScopes.length === 0` still records `recordCleanupDuration(..., 'success')` via `finally`. Operators see successful cleanup durations every cron
  tick for already-cleaned tenants, making the metric unreliable for alerting.                                                                                                  
   
  - **Partial scope deletion not counted in metrics**                                                                                                                           
    `tenant-delete.service.ts:88-104`                                                                                                                                         
    When `failedFolders.length > 0`, `recordCleanupScopesDeleted` is not called, so successfully deleted sub-folders in a partial failure are silently dropped from the metric.
                                                                                                                                                                                
  - **Root scope preservation is undocumented**
    `tenant-delete.service.ts:42-49`                                                                                                                                            
    Only child scopes are deleted; the root scope is intentionally preserved. A comment explaining _why_ would prevent a future developer from "fixing" this.                   
   
  ---                                                                                                                                                                           
                                                                                                                                                                              
  ### 🔄 Consistency

  - **`forEach` vs `for...of` inconsistency**                                                                                                                                   
    `files.service.ts:360` vs `:393`
    `getContentIdsByScope` uses `for...of` while the sibling `getIdsByScopeAndMetadataKey` uses `.forEach()`. Align to `for...of`.                                              
                                                                                                                                                                                
  - **Cleanup counters missing `result` label**
    `metrics.service.ts:158-164`                                                                                                                                                
    Every other counter uses `result: 'success' | 'failure'`. The new cleanup counters omit it, making alerting inconsistent.                                                 
                                                                                                                                                                                
  - **Missing `beforeEach(vi.clearAllMocks)` in delete service spec**
    `tenant-delete.service.spec.ts`                                                                                                                                             
    The module-level `mockLogger` is shared across tests without clearing between them, unlike every other spec file.                                                         
                                                                                                                                                                                
  ---
                                                                                                                                                                                
  ### 🧪 Test Gaps                                                                                                                                                            

  - [x] No test for the failure-path metric (`recordCleanupDuration` with `'failure'`)
  - [x] No test asserting scope deletion is attempted for scopes where content deletion failed (current test passes by coincidence)
  - [x] No test for large result sets from `getContentIdsByScope` (pagination boundary)                                                                                         
  - [x] No negative assertion on `recordCleanupScopesDeleted` in the partial failure test

services/confluence-connector/src/scheduler/tenant-sync.scheduler.ts
```diff
@@ -42,6 +33,24 @@ export class TenantSyncScheduler implements OnModuleInit, OnModuleDestroy {
...
+      return;
+    }
+
+    if (this.tenantRegistry.tenantCount === 0) {
+      this.logger.warn({ msg: 'No tenants registered — no jobs will be scheduled' });
+      return;
+    }
+
+    for (const tenant of this.tenantRegistry.getAllTenants()) {
+      this.logger.log({ tenantName: tenant.name, msg: 'Triggering initial job' });
```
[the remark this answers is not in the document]
lorand93: logic was because it's no longer just a sync as it might be a delete /cleanup also. 
Left it on sync:
```
      this.logger.log({ tenantName: tenant.name, msg: 'Triggering first sync on startup' });
```

packages/unique-api/src/files/files.queries.ts
```diff
@@ -176,6 +176,15 @@ export const CONTENT_ID_BY_SCOPE_AND_METADATA_KEY = gql`
     }
 `;
 
+export interface ContentByScopeInput {
```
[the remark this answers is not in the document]
lorand93: yup, I saw this also but didn't make the change.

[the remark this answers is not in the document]
lorand93: adjusted now

services/confluence-connector/src/config/tenant-config-loader.ts
```diff
@@ -36,11 +36,13 @@ export type TenantConfig = z.infer<typeof TenantConfigSchema>;
 export interface NamedTenantConfig {
   name: string;
   config: TenantConfig;
+  status: 'active' | 'deleted';
```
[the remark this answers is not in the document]
lorand93: exported the enum and introduced LoadedTenantStatus type (union of Active | Deleted)

services/confluence-connector/src/tenant/tenant-delete.service.ts
```diff
@@ -0,0 +1,124 @@
...
+      this.metrics.recordCleanupDuration(elapsedSeconds(startTime), 'success');
+    } catch (error) {
+      this.metrics.recordCleanupDuration(elapsedSeconds(startTime), 'failure');
+      throw error;
+    } finally {
+      tenant.isScanning = false;
+    }
+  }
+
+  private async deleteContentByScopes(scopes: Scope[]): Promise<void> {
```
Ostrzy: Opus raised a pretty good point, that if every single call in `deleteContentByScopes` and `deleteChildScopes` fails, we will still report it as success, no? 🤔 Shouldn't we maybe gather result and if we had any failures, throw or something? 🤔

services/confluence-connector/src/tenant/tenant-delete.service.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (childScopes.length === 0) {
+        this.logger.log({ tenantName: this.tenantName, msg: 'Already cleaned up, skipping' });
+        return;
+      }
+
+      // Two-step deletion is intentional: deleteContentByScopes removes files in batches per
+      // scope, then deleteChildScopes with recursive:true catches anything that failed in the
+      // first step. This ensures no content is orphaned even if individual file deletions fail.
+      await this.deleteContentByScopes(childScopes);
+      await this.deleteChildScopes(childScopes);
```
Ostrzy: Shouldn't we also clear the `externalId`? 🤔 So that user can remove the scope if they want 🤔 And we have the signal that deletion happened as planned 🤔

services/confluence-connector/src/tenant/tenant-delete.service.ts
```diff
@@ -0,0 +1,124 @@
...
+      });
+      return;
+    }
+
+    tenant.isScanning = true;
+    const startTime = Date.now();
+
+    try {
+      const rootScope = await this.uniqueClient.scopes.getById(this.scopeId);
+      if (!rootScope) {
```
Ostrzy, line 34: We should probably have another check here that if the root scope have no `externalId`, that means we already removed all the content, no? 🤔 And we then have to be sure all content is deleted 😄

services/confluence-connector/src/tenant/tenant-delete.service.ts
```diff
@@ -0,0 +1,124 @@
...
+      tenant.isScanning = false;
+    }
+  }
+
+  private async deleteContentByScopes(scopes: Scope[]): Promise<void> {
+    for (const scope of scopes) {
+      try {
+        const contentIds = await this.uniqueClient.files.getContentIdsByScope(scope.id);
+        if (contentIds.length > 0) {
+          await this.uniqueClient.files.deleteByIds(contentIds);
```
Ostrzy: `deleteByIds` doesn't return any info on partial successes that we could use for metrics? 🤔

Ostrzy (approved):

packages/unique-api/src/files/__tests__/files.service.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+function createBatchResponse(ids: string[]): ContentByScopeResult {
+  return {
+    paginatedContent: {
+      nodes: ids.map((id) => ({ id })),
+      totalCount: ids.length,
+    },
+  };
+}
+
+describe('FilesService', () => {
```
nicolae536, line 36: I think we can remove this file -> it literally tests nothing we are calling an external api and returning a zod parsed response - typescript already tests this

packages/unique-api/src/files/files.service.ts
```diff
@@ -350,8 +375,8 @@ export class FilesService implements UniqueFilesFacade {
     const ids: string[] = [];
     while (hasMore) {
       const batchResult = await this.ingestionClient.request<
-        ContentByScopeAndMetadataKeyResult,
-        ContentByScopeAndMetadataKeyInput
+        ContentByScopeResult,
```
nicolae536: Hmm this was renamed by the query still returns the same thing

packages/unique-api/src/files/files.service.ts
```diff
@@ -340,6 +340,31 @@ export class FilesService implements UniqueFilesFacade {
     return successCount;
   }
 
+  public async getContentIdsByScope(scopeId: string): Promise<string[]> {
+    let hasMore = true;
+    let skip = 0;
+    const ids: string[] = [];
+    while (hasMore) {
```
nicolae536, line 348: this while repets n times at some point we should make a utility which just calls for more but for now it's fine

services/confluence-connector/src/tenant/tenant-delete.service.ts
lorand93: ✅ bbda496d0b5df5a5c9ec279df61cc29cebe74add

services/confluence-connector/src/tenant/tenant-delete.service.ts
lorand93: absolutely true. Added that in f5e2a19749e54e474e4479993dce353bf86d3db3

services/confluence-connector/src/tenant/tenant-delete.service.ts
lorand93: 23fe86b86d44d6194b39ad0b20e2c811d1c03219

lorand93: added it in.

services/confluence-connector/src/tenant/tenant-delete.service.ts
lorand93: yes indeed, I implemented this now with opus and remove the existing todo comment

packages/unique-api/src/files/__tests__/files.service.spec.ts
lorand93: tests 2 and 3 cover the pagination loop — specifically the skip offset accumulation and the full-batch-then-empty-batch termination case. I shall trim the "toHaveBeenCalledWith" assertion in test 1 as it is redundant.

✅

packages/unique-api/src/files/files.service.ts
lorand93: renamed it to something more generic: PaginatedContentIdsQueryResult PaginatedContentIdsQueryInput

lorand93:
bugbot review

PR #403, opened 2026-03-25
docs(confluence-connector,ci): add public documentation

lorand93:
## Summary

Add public documentation for the Confluence Connector, structured to mirror the SharePoint Connector documentation template.

### Documentation structure

```
services/confluence-connector/docs/
├── README.md                        Overview and feature summary
├── faq.md                           Frequently asked questions
├── operator/
│   ├── README.md                    Operator guide, deployment checklist, network requirements
│   ├── authentication.md            OAuth 2LO, PAT, Zitadel, secret resolution, token flows
│   ├── configuration.md             Environment variables, tenant YAML, scheduling, logging
│   └── deployment.md                Container image, Helm chart, Terraform, upgrading
└── technical/
    ├── README.md                    Key concepts and documentation index
    ├── architecture.md              System components, connectivity, hosting models
    ├── flows.md                     Content sync, discovery, file diff, ingestion, error handling
    ├── permissions.md               Confluence and Unique API permissions
    └── security.md                  Security principles, data handling, compliance
```

### Deferred

- Unique API metrics and custom connector metrics tables in configuration.md — will be added after merging the observability-metrics PR

## Test plan

- [x] All internal markdown links resolve correctly
- [x] Content accuracy reviewed against current codebase

Ref: UN-16966


Ostrzy:
## Documentation Review — Correctness, Clarity & Consistency

Reviewed all 11 doc files against the actual source code, Helm charts, and the working SharePoint connector docs (as a reference for Confluence publishing conventions).

**Totals: 6 high, 8 medium, 12 low**

---

### 🔴 High Severity

#### 1. All 35 cross-doc anchor links use wrong case convention
Every anchor link uses lowercase kebab-case (`#file-diff-mechanism`) which will **break on Confluence**. The SharePoint docs (already published and working) use Title-Case (`#File-Diff-Mechanism`). All 35 links need converting.

Examples:
| Current (broken) | Should be |
|---|---|
| `#file-diff-mechanism` | `#File-Diff-Mechanism` |
| `#discovery-phase` | `#Discovery-Phase` |
| `#secret-resolution` | `#Secret-Resolution` |
| `#multi-tenancy-support` | `#Multi-Tenancy-Support` |

#### 2. Broken anchor: `flows.md#safety-checks` — target heading does not exist
- `README.md:62`
- `faq.md:186`

No "Safety Checks" heading exists anywhere in `flows.md`.

#### 3. Broken anchor: `architecture.md#multi-tenancy-model` — wrong heading name
- `README.md:94`
- `faq.md:253`

Actual heading is `## Multi-Tenancy Support` (`#multi-tenancy-support`). Note that `configuration.md:33` already uses the correct one.

#### 4. DC OAuth setup lists Cloud-style scopes (`authentication.md:113-125`)
Option B (Data Center OAuth) lists granular Cloud scopes (`search:confluence`, `read:confluence-content.all`, etc.) but Data Center uses a single `READ` scope. Confirmed by:
- Code: `oauth2lo-auth.strategy.ts:20` → `const DC_TOKEN_SCOPE = 'READ'`
- `permissions.md:15` → "OAuth 2.0 application configured with `READ` scope (instance-wide)"

#### 5. DC OAuth references wrong admin UI (`authentication.md:110-111`)
Points to `admin.atlassian.com` (Cloud-only Atlassian Admin Console) for DC setup. Data Center OAuth application links are configured in the DC instance's own admin interface.

#### 6. Missing `unique.ingestionConfig` field in `configuration.md`
This config option exists in the schema, Helm template, and Helm values, but is completely absent from the configuration docs.

---

### 🟡 Medium Severity

#### 7. Cloud OAuth scope contradiction (`authentication.md` vs `permissions.md`)
`authentication.md` lists 7 explicit scopes to configure on the service account. `permissions.md` says "no explicit OAuth scopes are sent in the token request." Both may be true (scopes granted on the account vs. sent in the request body), but this needs clarification.

#### 8. Resource table misleading (`operator/README.md:87-88`)
Says Memory "1 Gi minimum allocation" but actual `requests.memory` is `512Mi` (`values.yaml:47`); 1Gi is the limit. CPU "1 core" doesn't mention there's no CPU limit set.

Suggested: `| **Memory** | 512 Mi request / 1 Gi limit | Default Helm allocation |`

#### 9. MAX_HEAP_MB shown as 1920 without context (`faq.md:313`)
`configuration.md` and `architecture.md` distinguish Helm (1920) vs Docker (1024) defaults, but FAQ only shows 1920.

#### 10. Labels `ai-ingest`/`ai-ingest-all` implied as defaults (`README.md:54-56`)
The code schema (`confluence.schema.ts:33-34`) defines these as `z.string()` with no `.default()` — they are required fields. The docs present them as if they're default values. Should use "e.g." framing.

#### 11. "Recursive/hierarchical ingestion mode" listed as unsupported (`README.md:225`)
Easily confused with `ai-ingest-all` descendant discovery which IS supported. The actual limitation is flat scope structure (no nested sub-scopes mirroring Confluence page hierarchy).

Suggested: "Hierarchical scope structure (all pages from a space are placed in a single flat scope; sub-scopes mirroring the Confluence page tree are not created)"

#### 12. Contradictory phrasing about skipped-type descendants (`README.md:77`, `faq.md:48`)
"Descendants of all non-skipped types (including skipped ones like databases)" contradicts itself. Intent: when a skipped parent has the all-descendants label, its children are still ingested; only the parent is excluded.

Suggested: "When a skipped content type (such as a database) carries the all-descendants label, its child pages are still discovered and ingested; only the skipped parent itself is excluded."

#### 13. Attachment pagination differs by deployment type (`flows.md:197`)
Cloud uses v2 API (`/wiki/api/v2/pages/{pageId}/attachments`) because v1 pagination returns 410 Gone. Data Center uses v1 `_links.next` pagination. Docs don't distinguish.

#### 14. MAX_HEAP_MB confusing dual-row presentation (`architecture.md:271-274`)
Lists "Default max heap size: 1024 MB" and "Helm chart MAX_HEAP_MB: 1920 MB" as separate rows. The 1024 value isn't set anywhere in the codebase or Helm chart.

---

### 🟢 Low Severity

#### 15. Boundary condition: "smaller than" vs "at most" (`faq.md:211`)
Says "smaller than `maxFileSizeMb`" but code uses `>` check (`confluence-page-scanner.ts:143`), meaning files exactly at the limit ARE allowed. Should be "at most".

#### 16. Inconsistent external URL for DC REST API docs
`permissions.md` uses `developer.atlassian.com/server/confluence/rest/`, all other files use `docs.atlassian.com/ConfluenceServer/rest/latest/`. Pick one.

#### 17. Ambiguous OAuth/PAT version boundary (`architecture.md:92`)
"OAuth 2.0 (Cloud and Data Center) or PAT (Data Center below 10.1)" doesn't clarify that OAuth 2LO requires DC 10.1+.

#### 18. "No TLS downgrade" claim incomplete (`security.md:91`)
True for connector config, but doesn't warn about the `NODE_TLS_REJECT_UNAUTHORIZED=0` env var which can bypass TLS verification at runtime.

#### 19. Missing "no CPU limit" row (`architecture.md:268`)
Resource table lists CPU request but doesn't mention there's no CPU limit (intentionally burstable). Could surprise operators.

#### 20. Hosting Models section duplicated verbatim
`authentication.md:366-432` and `architecture.md:195-262` contain identical content. Drift risk — should live in one place with a cross-reference.

#### 21. Egress domain list duplicated in 3+ places
Appears in `operator/README.md`, `authentication.md`, and `deployment.md`. Already required a fix across all locations in commit `6d59363a`. Consider a single canonical list with cross-references.

#### 22. `allowedExtensions` default missing `xlsx` (`values.schema.json:403`)
Code default includes `xlsx`: `['pdf', 'docx', 'xlsx', 'ppt', 'pptx', 'txt', 'csv', 'html']`, but schema default omits it.

#### 23. Dead env var: `MAX_FILE_SIZE_MB` in Helm values (`values.yaml:9`)
Set in the Helm chart defaults but never read by application code. Max file size comes from tenant YAML config (`ingestion.attachments.maxFileSizeMb`).

#### 24. Folder skip mechanism not explained (`README.md:77`)
Folders pass discovery (appear in file diff) but are skipped at ingestion due to empty body. This differs from other skipped types (databases, whiteboards, embeds) which are filtered during discovery.

#### 25. "recommended: `ai-ingest`" misleading (`faq.md:36-37`)
"Recommended" implies optional with a suggested value. These are required fields. Should be "e.g." or "convention:".

#### 26. Backend-service subchart version not mentioned (`deployment.md:25`)
The chart depends on `backend-service ~10.1.0` (from `Chart.yaml`). Useful for operators to know.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

lorand93:
I removed it from services/confluence-connector/docs/operator/authentication.md and just refferenced it instead. I think SPC also has this duplication which might be intentional, i'm fine in any way, less things to update when it changes
>20. Hosting Models section duplicated verbatim
authentication.md:366-432 and architecture.md:195-262 contain identical content. Drift risk — should live in one place with a cross-reference.

lorand93:
**Attachment pagination differs by deployment type (flows.md)**
The current text at line 199 says "additional attachments are fetched via pagination" but doesn't mention that Cloud uses v2 API while DC
  uses v1 _links.next. However, line 229 in the DC table already notes "Uses v1 pagination (_links.next)" for DC attachments.

  The fix would be to expand line 199:

  Current:
  > If a page has more than 25 attachments (the Confluence inline limit), additional attachments are fetched via pagination.

  Proposed:
  > If a page has more than 25 attachments (the Confluence inline limit), additional attachments are fetched via pagination. Cloud uses the v2
   attachments endpoint; Data Center uses v1 pagination via `_links.next`.
   
As this falls into the **Implementation Detail** bucket I won't do the change for now.

lorand93:
Egress domain list duplicated in 3+ places
 
 The egress domain list appears in `operator/README.md` (Network Requirements table), `operator/deployment.md` (troubleshooting section), and `operator/authentication.md` (troubleshooting section). The lists are consistent in content but duplicated.
 The fix would be to keep the canonical list in `operator/README.md` and replace the other occurrences with a cross-reference like "See the network requirements table in the Operator Guide for the full list of required endpoints." (deployment.md line 253 already does this).
 
 Since the duplication serves readability in troubleshooting sections (operators shouldn't have to jump to another page mid-diagnosis), I won't do this change for now unless indicated else

services/confluence-connector/docs/operator/authentication.md
```diff
@@ -0,0 +1,436 @@
...
+    mode: oauth_2lo
+    clientId: your-oauth-client-id
+    clientSecret: os.environ/CONFLUENCE_CLIENT_SECRET
+```
+
+The `clientSecret` field uses the `os.environ/` prefix to resolve the value from an environment variable at runtime (see [Secret Resolution](#Secret-Resolution)).
+
+#### Option B: OAuth 2.0 (2LO) -- Data Center
+
+1. In the [Atlassian Admin Console](https://admin.atlassian.com/), go to **Settings** > **User Management** > **Service Accounts**.
```
Ostrzy: This link is most certainly incorrect for data center 😛

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+| `updatedAt` | Last modification timestamp from Confluence | Change detection |
+
+### Safety Checks
+
+After each space's file diff is computed, two guards run to prevent accidental full deletion of content. Both abort the current tenant sync cycle if triggered.
+
+| Guard | Trigger Condition | Rationale |
+|---|---|---|
+| **Zero-submission guard** | Discovery submitted 0 items but the diff contains deletions | Likely a bug in page fetching or an authentication failure that returned no results. Deleting all stored content would be destructive. |
+| **Full-deletion guard** | The number of files to delete equals the total number of files stored in Unique for that space | Likely a key format change (e.g., toggling `useV1KeyFormat`) that causes all existing keys to appear unrecognized. |
```
Ostrzy: Isn't the check more nuanced now, with the check if the values in the key overlap as well? 🤔

services/confluence-connector/docs/operator/authentication.md
```diff
@@ -0,0 +1,436 @@
...
+This user identity is referenced:
+
+- In `cluster_local` mode: as the `x-user-id` header value in `serviceExtraHeaders`. This **must** be the ID of an actual service user in Zitadel. It cannot be an arbitrary value.
+- In `external` mode: implicitly via the Zitadel client credentials (`zitadelClientId` / `zitadelClientSecret`).
+
+Steps:
+
+1. Navigate to Zitadel.
+2. Create a new service user.
+3. For `external` mode: assign the required authorizations listed above. For `cluster_local` mode: no authorizations are needed.
```
Ostrzy: Don't we also need the client secret from Zitadel for the service user? 🤔 Or this is somehow covered in the linked PUBDOC guides?

services/confluence-connector/docs/operator/authentication.md
```diff
@@ -0,0 +1,436 @@
...
+
+For detailed instructions on creating and configuring a service user, see:
+- [How To Configure A Service User](https://unique-ch.atlassian.net/wiki/spaces/PUBDOC/pages/1411023075/How+To+Configure+A+Service+User)
+- [Understand Roles and Permissions](https://unique-ch.atlassian.net/wiki/spaces/PUBDOC/pages/1411023168)
+
+### 2. Create the Root Scope in Unique
+
+The connector requires a pre-existing root scope in Unique. The root scope ID is configured in the tenant YAML under `ingestion.scopeId`. If the scope does not exist at startup, the connector fails.
+
+At startup, the connector automatically grants itself access on the root scope and will create child scopes for each Confluence space that has content to ingest. See the [Scope Hierarchy](../technical/flows.md#Scope-Hierarchy) for details.
```
Ostrzy, line 72: Don't we need to grant access for service user via GraphQL API when we're in external mode? 🤔 For `cluster_local` for sure we don't we can simply add the permissions, but for `external` I think we need to be assigned 🤔

services/confluence-connector/docs/operator/configuration.md
```diff
@@ -0,0 +1,411 @@
...
+
+| Field | Required | Default | Description |
+|---|---|---|---|
+| `instanceType` | Yes | -- | `cloud` or `data-center` |
+| `baseUrl` | Yes | -- | Base URL of the Confluence instance (e.g., `https://acme.atlassian.net`). Must not end with a trailing slash |
+| `cloudId` | Yes (Cloud only) | -- | Atlassian Cloud ID (UUID) for the Confluence site |
+| `auth` | Yes | -- | Authentication configuration (see [Authentication](./authentication.md#Confluence-Authentication-Methods)) |
+| `apiRateLimitPerMinute` | Yes | -- | Number of Confluence API requests allowed per minute |
+| `ingestSingleLabel` | Yes | -- | Confluence label that marks individual pages for synchronization (e.g., `ai-ingest`) |
+| `ingestAllLabel` | Yes | -- | Confluence label that marks a page and all its descendants for synchronization (e.g., `ai-ingest-all`) |
```
Ostrzy: I think it would be nice to format tables in markdown as well so it's easier to read / review 🙂

services/confluence-connector/docs/operator/configuration.md
```diff
@@ -0,0 +1,411 @@
...
+
+| Level | Description |
+|---|---|
+| `fatal` | Unrecoverable errors |
+| `error` | Error conditions |
+| `warn` | Warning conditions |
+| `info` | General operational information (default) |
+| `debug` | Detailed debugging information |
+| `trace` | Very detailed trace-level information |
+| `silent` | No logging output |
```
Ostrzy: We do not use silent and fatal. I don't think we even use trace at all 🤔

services/confluence-connector/docs/operator/configuration.md
```diff
@@ -0,0 +1,411 @@
...
+        #   severity: "critical"
+        #   threshold: 0.05
+    additionalLabels: {}
+      # environment: production
+      # team: backend
+```
+
+Alerts require the `monitoring.coreos.com/v1` API (Prometheus Operator) to be available in the cluster.
+
+## Complete Re-ingestion
```
Ostrzy, line 371: Can't you mark tenant as deleted, and mark back as active? 🤔

services/confluence-connector/docs/technical/architecture.md
```diff
@@ -0,0 +1,292 @@
...
+- Zitadel token validation is not needed
+- Services communicate within the cluster via internal HTTP
+- Company and user scope is maintained via request headers:
+  - `x-company-id`
+  - `x-user-id`
+- Upload URLs are rewritten to the ingestion service's scoped upload endpoint to avoid hairpinning through the gateway
+
+## Multi-Tenancy Support
+
+Multiple Confluence instances (tenants) can be configured in a single deployment. Each tenant is isolated at runtime and has its own service instances, API clients, authentication strategy, and sync schedule. All per-tenant services are initialized at startup.
```
Ostrzy: When we're talking about Kubernetes and deployments it can be understood that we spawn a separate pod per tenant or something 🤔

services/confluence-connector/docs/faq.md
```diff
@@ -0,0 +1,330 @@
...
+- Operates as a background service without user interaction
+- Supports both Confluence Cloud and Confluence Data Center
+
+### How does this differ from the Confluence Connector v1?
+
+**Answer:**
+
+| Aspect | v1 | v2 |
+|---|---|---|
+| Architecture | Monorepo module | Standalone service |
```
Ostrzy: That seems irrelevant? 🤔

services/confluence-connector/docs/faq.md
```diff
@@ -0,0 +1,330 @@
...
+| Deployment | Part of a larger Node.js application | Independent Kubernetes container |
+| Multi-tenancy | Not supported | Multiple Confluence instances in a single deployment |
+| Attachment ingestion | Not supported | Supported with configurable extensions and size limits |
+| Change detection | Re-ingests all content | File-diff mechanism ingests only new or modified items |
```
Ostrzy: I am pretty sure that is simply not true 😛

services/confluence-connector/docs/faq.md
```diff
@@ -0,0 +1,330 @@
...
+
+### What happens to attachments when their parent page is unlabeled?
+
+**Answer:** Attachments are discovered as children of labeled pages. If a page is no longer discovered (because its label was removed or the page was deleted), its attachments are also missing from the discovery results and are deleted from Unique via the file diff mechanism.
+
+### How are scopes organized in Unique?
+
+**Answer:** Scopes follow a two-level hierarchy: a root scope configured per tenant, and child scopes automatically created for each Confluence space key. Child scopes inherit access from the root scope. See the [Scope Hierarchy](./technical/flows.md#Scope-Hierarchy) for details.
+
+### What is the ingestion key format?
```
Ostrzy: It seems to be very technical for the FAQ 😛

services/confluence-connector/docs/faq.md
```diff
@@ -0,0 +1,330 @@
...
+
+**Answer:** Yes. Each Confluence instance is configured as a separate tenant with its own YAML configuration file. All tenants run within a single connector deployment with independent authentication, API clients, and sync schedules. See [Architecture -- Multi-Tenancy Support](./technical/architecture.md#Multi-Tenancy-Support) for details on tenant isolation and per-tenant service instances.
+
+### How do I add a new tenant?
+
+**Answer:** Create a new YAML configuration file following the naming convention `<tenant-name>-tenant-config.yaml` in the directory matched by the `TENANT_CONFIG_PATH_PATTERN` environment variable. The connector must be restarted to pick up new tenant configuration files.
+
+### Can two tenants use the same scope ID?
+
+**Answer:** This is not recommended. Each tenant should have its own root scope to avoid conflicts in scope ownership and content management.
```
Ostrzy: Don't we literally fail if someone does that? 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+### Overview
+
+```mermaid
+flowchart TB
+    subgraph Trigger["Trigger"]
+        Scheduler["Scheduler<br/>(cron)"]
+    end
+
+    subgraph Initialization["Initialization"]
+        Guard["Re-entrancy guard"]
```
Ostrzy: The hell is `Re-entrancy guard`? 😄

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+        SearchCQL["CQL label search"]
+        Descendants["Fetch descendants<br/>(ingestAllLabel pages)"]
+        Deduplicate["Deduplicate pages"]
+        FilterTypes["Filter skipped content types"]
+        ExtractAttachments["Extract attachments<br/>(if enabled)"]
+    end
+
+    subgraph Processing["Processing Phase"]
+        Diff["Per-space file diff<br/>(Unique Platform)"]
+        Validate["Safety checks"]
```
Ostrzy: Maybe something slightly more like "Redundant re-ingestion guards" or something like that? 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+    Connector->>Unique: Initialize root scope<br/>(grant access, verify exists)
+
+    rect rgb(200, 210, 240)
+        Note over Connector,Confluence: Discovery Phase
+        Connector->>Confluence: CQL search: pages with<br/>ingestSingleLabel OR ingestAllLabel
+        Confluence->>Connector: Labeled pages<br/>(with inline attachments)
+
+        opt Pages with ingestAllLabel
+            Connector->>Confluence: CQL ancestor query<br/>(batches of 100 root IDs)
+            Confluence->>Connector: Descendant pages
```
Ostrzy: Do these have inline attachments as well? If we mark it in the previous call then I guess both or none 😄

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+
+## Discovery Phase
+
+Pages are discovered through a CQL-based label search, then attachments are optionally extracted from the already-fetched page objects.
+
+**Important:** CQL search results are filtered by the authenticated user's permissions. Pages in spaces the service account cannot access are silently excluded from results. On Data Center, the service account's space access can optionally be restricted to specific spaces. When no spaces are specified, the service account has read access to all spaces. If space restrictions are configured, pages in excluded spaces never appear in discovery results and no error is produced.
+
+### Discovery Sequence
+
+```mermaid
```
Ostrzy: Does this chart bring anything? I feel like it's basically a full repeat of the first chart on this page 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+Pages labeled with `ingestAllLabel` trigger a descendant search:
+
+1. Collect all page IDs that carry the `ingestAllLabel`
+2. Batch IDs into groups of 100
+3. For each batch, execute CQL: `ancestor IN ({batch}) AND type != attachment`
+4. Deduplicate results with labeled pages by page ID
+
+### Attachment Extraction
+
+When `attachments.mode` is `enabled` (default), attachments are extracted from the already-fetched page objects -- no additional API calls are made during extraction. Attachments were inlined by the `expand=children.attachment` parameter during the CQL search.
```
Ostrzy: Isn't it untrue if there are more than 25 attachmants? 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+| Embed / Smart Link | **No** | No (only has `embedUrl`) | Explicitly skipped. Only contains a URL reference, no renderable body. |
+| Folder | **No** (effectively) | No (organizational container) | Not explicitly skipped, but has no body -- skipped by the empty-body filter. Descendants are still discovered. |
+| Comment (inline/footer) | **No** | Yes (`body.storage` / ADF) | Not discovered -- comments do not appear in label/ancestor CQL results. |
+| Live Doc (page subtype) | **Yes** (as page) | Yes (`body.storage` / ADF) | Subtype of page. Passes through as a regular page. |
+| Custom Content (app-defined) | **No** | Yes (`body.storage`) | Not discovered -- uses `ac:key:type` format, not matched by standard CQL. |
+| Task (standalone) | **No** | Yes (`body.storage` / ADF) | Not a CQL-searchable content type. Only accessible via `/tasks` v2 endpoint. |
+
+#### Confluence Data Center
+
+| Content Type | Exists in DC? | Ingested? | Notes |
```
Ostrzy: This Exitsts in DC is weird, we should drop it and instead just add a text below the table that Whiteboard, Database, Embed, Folder, Live Doc are not available in Data Center 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+| Custom Content (plugin-defined) | Yes | **No** | Accessed via plugin-specific REST APIs, not standard `/rest/api/content`. |
+| Whiteboard, Database, Embed, Folder, Live Doc | No | N/A | Cloud-only features. Do not exist in Data Center. |
+
+## File Diff Mechanism
+
+The connector uses the Unique platform's server-side file diff API (`/v2/content/file-diff`) to detect changes between sync cycles. The connector does **not** compute local content hashes -- instead, it sends each item's `key`, `url`, and `updatedAt` timestamp to the diff endpoint, which returns categorized results. The diff is called once per Confluence space.
+
+### State Comparison
+
+```mermaid
```
Ostrzy: Maybe it's me but I have no clue what this chart brings 🤷🏼 Especially with the Change Detection Logic chart that is more detailed

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+After each space's file diff is computed, two guards run to prevent accidental full deletion of content. Both abort the current tenant sync cycle if triggered.
+
+| Guard | Trigger Condition | Rationale |
+|---|---|---|
+| **Zero-submission guard** | Discovery submitted 0 items but the diff contains deletions | Likely a bug in page fetching or an authentication failure that returned no results. Deleting all stored content would be destructive. |
+| **Full-deletion guard** | The number of files to delete equals the total number of files stored in Unique for that space | Likely a key format change (e.g., toggling `useV1KeyFormat`) that causes all existing keys to appear unrecognized. |
+
+If neither guard triggers, deletions proceed normally. Partial deletions (removing some but not all files) are always allowed.
+
+To intentionally remove all content from a space, leave at least one page labeled for synchronization to avoid triggering these guards.
```
Ostrzy: Don't we have the deleted state for that that should be reversible? 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+
+The connector applies scenario-specific behavior to keep sync cycles stable:
+
+| Scenario | Typical Cause | Connector Behavior |
+|---|---|---|
+| Sync already in progress | Overlapping cron triggers or long-running sync | Skip the cycle entirely (re-entrancy guard) |
+| Root scope not found | Misconfigured `scopeId` or scope deleted | Abort the entire sync cycle |
+| Accidental full deletion detected | Bug in page fetching or key format change | Abort the entire tenant sync cycle |
+| Page fetch failure | Page deleted between discovery and content fetch, transient API error | Log error, skip the page, continue other pages |
+| Page not found | Page deleted between discovery and content fetch | Log warning, skip the page |
```
Ostrzy: How `Page fetch failure` is different from `Page not found`? 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+| Page with empty body | Page has no content (e.g., newly created) | Log, skip the page |
+| Attachment with zero bytes | Empty attachment | Log, skip the attachment |
+| Page ingestion failure | Upload error, registration error | Log error, skip the page |
+| Attachment ingestion failure | Download error, upload error | Destroy stream, log error, skip the attachment |
```
Ostrzy: Don't we also ensure that partial file is removed from KB? 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+| Content deletion failure | Unique API error | Log error, return 0 deleted count |
```
Ostrzy: I don't understand the `return 0 deleted count` 🤔

services/confluence-connector/docs/technical/flows.md
```diff
@@ -0,0 +1,392 @@
...
+| Unhandled sync error | Unexpected exception | Caught at top level, logged, sync state reset |
+
```
Ostrzy, line 311: Maybe worth mentioning somewhere here that is any error occurs during state fetching from Confluence, the sync is aborted to not submit partial results to file-diff as this can result in incurring unnecessary costs 🤔

Ostrzy (approved):

services/confluence-connector/docs/technical/flows.md
lorand93: prepared separate PR for this but updated it now here

services/confluence-connector/docs/operator/authentication.md
lorand93: I had two stages to this:
1. i checked initially and chat admin all seemed to be enough 
The connector self-grants access via createScopeAccesses at initialization (same as the SharePoint connector). With
  CHAT_ADMIN_ALL, this passes the @AllowedUserRoles guard. However, I checked the Gatekeeper role definitions and Chat_Admin
  does not include SCOPE_ACCESS permissions. If Gatekeeper is in enforce mode, the call would be denied.
  
2. when gatekeeper is enforced GATEKEEPER_RUNNING_MODE=enforce 
CHAT_ADMIN_ALL is NOT enough when Gatekeeper is in enforce mode. The Chat_Admin role definition does not include
  SCOPE_ACCESS permissions. The call passes the first guard (@AllowedUserRoles) but gets denied at the Gatekeeper step.

  The Gatekeeper has three modes:
  - disabled (default) - always allows
  - report_only - logs but allows
  - enforce - strictly enforces

lorand93: added comment to indicate what to do when gatekeeper is in enforced mode

services/confluence-connector/docs/operator/authentication.md
lorand93: it's worth being explicit here. added

services/confluence-connector/docs/operator/configuration.md
lorand93: added this instead of inactive

services/confluence-connector/docs/faq.md
lorand93: true

services/confluence-connector/docs/faq.md
lorand93: yes missed it

services/confluence-connector/docs/faq.md
lorand93: true. removing it.

services/confluence-connector/docs/faq.md
lorand93: now yes, when I added the docs initially no...

services/confluence-connector/docs/technical/flows.md
lorand93: reworded it as re-entryancy guard is an AI loved word.  "skip if sync in progress" is better

services/confluence-connector/docs/technical/flows.md
lorand93: Added "(with inline attachments)" to the descendant pages response to match the label search response

services/confluence-connector/docs/technical/flows.md
lorand93: 🤔 true  let's remove it

services/confluence-connector/docs/technical/flows.md
lorand93: corrected for more than 25

services/confluence-connector/docs/technical/flows.md
lorand93: Merged the two rows into a single "Page unavailable" scenario

services/confluence-connector/docs/technical/flows.md
lorand93: removed it, legit mistake

services/confluence-connector/docs/technical/flows.md
lorand93: true, added

services/confluence-connector/docs/technical/flows.md
lorand93: 👍

services/confluence-connector/docs/technical/flows.md
lorand93: added that in. Wasn't there because the implementation is in another pr

services/confluence-connector/docs/technical/flows.md
lorand93: Dropped the "Exists in DC?" column and moved the Cloud-only note to a line below the table.

services/confluence-connector/docs/technical/flows.md
lorand93: 👍

PR #434, opened 2026-04-06
feat(outlook-semantic-mcp,unique-api)!: Implement content expiration

nicolae536:
Refs: UN-19133, UN-16564

feat(outlook-semantic-mcp,unique-api)!: replaces ignoreBefore with expiration window

BREAKING CHANGE: Replace ignoredBefore date with retention window in days


services/outlook-semantic-mcp/src/unique/get-root-scope-path.ts
```diff
@@ -1,9 +1,9 @@
...
 
-export const getRootScopeExternalId = () => `/__Outlook MCP`;
+export const getRootScopeExternalId = () => `/Outlook MCP`;
 
 export const getRootScopePathForUser = (userProviderIdentifier: string): string =>
   `${getRootScopePath()}/${userProviderIdentifier}_uncategoried`;
 
 export const getRootScopeExternalIdForUser = (userProviderIdentifier: string): string =>
-  `Outlook_${userProviderIdentifier}_uncategoried`;
+  `OutlookMCP_${userProviderIdentifier}_uncategoried`;
```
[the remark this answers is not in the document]
nicolae536: Reverted -  I will handle the transition differently

services/outlook-semantic-mcp/docs/operator/configuration.md
```diff
@@ -160,7 +160,7 @@ mcpConfig:
       x-company-id: "<your-company-id>"
       x-user-id: "<your-zitadel-service-user-id>"
 
-  defaultMailFilters: '{"ignoredBefore":"2025-06-06","ignoredContents":[],"ignoredSenders":[]}'
+  defaultMailFilters: '{"retentionWindowInDays":365,"ignoredContents":[],"ignoredSenders":[]}'
```
Ostrzy: Why is the default different from the recommended values? 😛

services/outlook-semantic-mcp/docs/operator/deployment.md
```diff
@@ -140,11 +140,11 @@ mcpConfig:
...
-  # IMPORTANT: `ignoredBefore` is mandatory — the application will not start without it.
-  # Adjust it to control how far back the initial email sync goes. Setting this date far
-  # in the past can cause very large initial syncs for users with large mailboxes,
-  # potentially taking hours and consuming significant Microsoft Graph API quota.
-  defaultMailFilters: '{"ignoredBefore":"2025-06-06","ignoredContents":[],"ignoredSenders":[]}'
+  # IMPORTANT: `retentionWindowInDays` is mandatory — the application will not start without it.
+  # Adjust it to control how far back the initial email sync goes. Setting a large value can
+  # cause very large initial syncs for users with large mailboxes, potentially taking hours
+  # and consuming significant Microsoft Graph API quota.
+  defaultMailFilters: '{"retentionWindowInDays":365,"ignoredContents":[],"ignoredSenders":[]}'
```
Ostrzy: Again, value outside of recommended range 🙂

services/outlook-semantic-mcp/src/features/process-email/utils/should-skip-email.ts
```diff
@@ -25,9 +28,9 @@ export function shouldSkipEmail(
   context: { userProfileId: string },
 ): SkipResult {
   try {
-    if (filters.ignoredBefore && email.receivedDateTime) {
-      if (new Date(email.receivedDateTime) < filters.ignoredBefore) {
-        return { skip: true, reason: 'ignoredBefore' };
+    if (email.receivedDateTime) {
+      if (new Date(email.receivedDateTime) < computeIgnoredBefore(filters.retentionWindowInDays)) {
```
Ostrzy: Any reason for it not to be a single condition? 🤔

services/outlook-semantic-mcp/src/features/process-email/process-email.command.ts
```diff
@@ -232,6 +235,10 @@ export class ProcessEmailCommand {
       sourceKind: INGESTION_SOURCE_KIND,
       sourceName: INGESTION_SOURCE_NAME,
       storeInternally: this.configService.get('unique.storeInternally', { infer: true }),
+      expiresAt: new Date(
+        new Date(graphMessage.receivedDateTime).getTime() +
+          filters.retentionWindowInDays * 24 * 60 * 60 * 1000,
```
Ostrzy: But that is incorrect, isn't it? 🤔 When you do full sync, every mail will expired in, let's say, 90 days instead of mails from 2 months ago expiring in a month and only new mails expiring in 90 days 🤔

Ostrzy (changes requested):

Ostrzy:
Some good points in here, please take a look

Critical                                                                                                                                                                      
                                                                                                                                                                                
  1. Verify downstream GraphQL schema accepts expiresAt — The expiresAt field is added to ContentUpsertMutationInput and passed through registerContent / finalizeIngestion, but
   there's no evidence the server-side ContentCreateInput GraphQL type includes this field. If it doesn't, expiresAt will be silently dropped. No integration test covers this
  path.
  - :white_check_mark: Checked while testing it accepts it                  
                                                                                                                                                        
  2. No database migration for existing filters JSONB rows — Existing rows contain {"ignoredBefore": "2025-..."}, but the Zod schema now requires retentionWindowInDays:      
  z.number().int().positive(). Parsing old rows will crash. The SyncOnFilterChangeService.onModuleInit overwrites all rows at startup, which mitigates this if it runs before   
  any filter-parsing code — but if a queued AMQP message triggers shouldSkipEmail during startup, it will fail. A proper SQL data migration would be safer.
  - :white_check_mark: Not sure what to do here because if we change the env var the migration is not true anymore I would I would say let it crush since it has exponential backoff.
                                                                                                                                                                                
  Warnings                                                                                                                                                                      
  
  3. Rolling cutoff shifts mid-sync — silent data loss risk — computeIgnoredBefore() calls new Date() on every invocation. During a full sync spanning midnight UTC, the Graph  
  API query might fetch an email using cutoff April 7, but shouldSkipEmail later computes cutoff April 8 and rejects it. The email is fetched but discarded and may never be  
  retried. Fix: compute the cutoff once at sync-start and thread it through.
 - :white_check_mark: Intended behaviour if it expired meanwhile we should not take it.          
                                                                                           
  4. expiresAt not refreshed for already-ingested emails — expiresAt is only set during initial ingestion (uploadEmailForIngestion), not on re-sync or the metadata-updated / 
  content-unchanged paths. If the retention window is later increased, existing emails' expiresAt values won't be extended.        
- :white_check_mark:                     
                           
  5. computeIgnoredBefore is impure — flaky tests near midnight — The function and its tests both call new Date() independently. If the clock crosses midnight UTC between the
  two calls, assertions fail. Fix: accept an optional now?: Date parameter.    
 - :white_check_mark: Fixed                                                                                                 
  
  6. Docs recommend 95–180 days, defaults are 365 — The configuration guide says "values above 180 days are not recommended" but .env.example, values.yaml, Helm chart, and all
  examples ship with 365. This is contradictory.
 - :white_check_mark: Fixed
                                                                                                                                
  7. ignoredBefore schema changed from nullable to non-nullable — In GetFullSyncStatsResponse, ignoredBefore went from z.string().nullable() to z.string(). If external       
  consumers expect null, this is a breaking change.                
- :white_check_mark: Should have been always non nullable                                                                                                             
                                                                                                                                                                              
  Notes                                                                                                                                                                         
                                                                                                                                                                              
  8. Function name computeIgnoredBefore is stale — The concept is now "retention window cutoff," but the function still uses the old naming. Consider computeRetentionCutoff.   
  - :white_check_mark: Fixed
  
  10. tools.md is missing ignoredBefore from the documented response — The actual response schema includes both retentionWindowInDays and a computed ignoredBefore, but the docs
  only show retentionWindowInDays.         
  - :white_check_mark: Fixed                                                                                                                                     
  
  11. Comment in live-catch-up.command.ts:227 still says "ignoredBefore" — Stale comment: "The ignoredBefore check is applied in-memory below." Should reference the retention
  window instead.                                                                                    
  - :white_check_mark: Fixed                                                                           
  
  12. Pre-existing bug (not from this branch) — inbox-configuration-mail-filters.dto.ts:15: new RegExp(patternPart, patternPart) should be new RegExp(patternPart, flagsPart).
  Regex flags are being ignored.                                                                                      
  - :white_check_mark: Fixed

services/outlook-semantic-mcp/docs/operator/configuration.md
nicolae536: Fixed

services/outlook-semantic-mcp/docs/operator/deployment.md
nicolae536: Fixed

services/outlook-semantic-mcp/docs/faq.md
```diff
@@ -187,7 +187,7 @@ This must match exactly — including protocol, domain, and path — in both the
 
 ### What does `DEFAULT_MAIL_FILTERS` do?
 
-**Answer:** `DEFAULT_MAIL_FILTERS` is a JSON object that controls which emails are ingested during both full sync and live catch-up. It supports three filters: `ignoredBefore` (ISO 8601 date cutoff — required, the application will not start without it), `ignoredSenders` (RegExp patterns matching sender addresses), and `ignoredContents` (RegExp patterns matching subject or body).
+**Answer:** `DEFAULT_MAIL_FILTERS` is a JSON object that controls which emails are ingested during both full sync and live catch-up. It supports three filters: `retentionWindowInDays` (positive integer — required, the application will not start without it; emails older than this rolling window are excluded and stamped with an `expiresAt` date), `ignoredSenders` (RegExp patterns matching sender addresses), and `ignoredContents` (RegExp patterns matching subject or body).
```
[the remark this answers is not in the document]
nicolae536: Fixed

services/outlook-semantic-mcp/src/config/app.config.ts
```diff
@@ -23,7 +23,7 @@ const ConfigSchema = z
       .describe('The log level at which the services outputs (pino).'),
     selfUrl: stringToURL().describe('The URL of the MCP Server. Used for OAuth callbacks.'),
     defaultMailFilters: json(inboxConfigurationMailFilters).describe(
-      'Default mail filters applied when syncing emails (e.g. {"ignoredBefore":"2024-01-01", "ignoredSenders": [], "ignoredContents": [] }). ',
+      'Default mail filters applied when syncing emails (e.g. {"retentionWindowInDays":365, "ignoredSenders": [], "ignoredContents": [] }). ',
```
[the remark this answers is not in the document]
nicolae536: Fixed

services/outlook-semantic-mcp/src/utils/date/__tests/compute-retention-cutoff-date.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { describe, expect, it } from 'vitest';
+import { computeIgnoredBefore } from './inbox-configuration-mail-filters.dto';
+
+describe('computeIgnoredBefore', () => {
+  it('returns UTC midnight of the date that is N days ago', () => {
```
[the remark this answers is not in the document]
nicolae536: Added the checks

services/outlook-semantic-mcp/.env.example
```diff
@@ -1,6 +1,6 @@
 LOG_LEVEL=debug
 SELF_URL=http://localhost:9542
-DEFAULT_MAIL_FILTERS={"ignoredBefore":"2025-06-06", "ignoredContents": [], "ignoredSenders": []} # JSON string of default mail filters applied when syncing emails (e.g. ignoredBefore limits sync to emails on or after the given date)
+DEFAULT_MAIL_FILTERS={"retentionWindowInDays":365, "ignoredContents": [], "ignoredSenders": []} # JSON string of default mail filters applied when syncing emails (e.g. retentionWindowInDays limits sync to emails received within the last N days, rolling forward daily)
```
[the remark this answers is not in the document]
nicolae536: Fixed

services/outlook-semantic-mcp/src/features/process-email/utils/should-skip-email.ts
nicolae536: Fixed

nicolae536:
For migration of QA / Prod I plan to execute a query in the ingestion db to set the dates correctly 🤔

services/outlook-semantic-mcp/src/db/schema/inbox/inbox-configuration-mail-filters.dto.ts
```diff
@@ -28,18 +28,25 @@ const regexPattern = z.string().transform((pattern, ctx) => {
...
 export const inboxConfigurationMailFilters = z.object({
-  ignoredBefore: z.coerce.date(),
+  retentionWindowInDays: z.number().int().positive(),
   ignoredSenders: z.array(regexPattern).optional().default([]),
   ignoredContents: z.array(regexPattern).optional().default([]),
 });
 
 export type InboxConfigurationMailFilters = z.infer<typeof inboxConfigurationMailFilters>;
 
+export function computeIgnoredBefore(days: number): Date {
```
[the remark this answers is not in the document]
nicolae536: Renamed

services/outlook-semantic-mcp/src/features/process-email/process-email.command.ts
```diff
@@ -232,6 +235,10 @@ export class ProcessEmailCommand {
       sourceKind: INGESTION_SOURCE_KIND,
       sourceName: INGESTION_SOURCE_NAME,
       storeInternally: this.configService.get('unique.storeInternally', { infer: true }),
+      expiresAt: new Date(
+        new Date(graphMessage.receivedDateTime).getTime() +
```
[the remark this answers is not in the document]
nicolae536: I made it strict since in zod schema was strict - I also checked `drafts` because I thought they do not have this date - I never saw a message without this date

packages/unique-api/src/ingestion/ingestion.queries.ts
```diff
@@ -79,29 +79,38 @@ export const CONTENT_UPSERT_MUTATION = gql`
   }
 `;
 
-export const CONTENT_UPDATE_METADATA_MUTATION = gql`
-  mutation ContentUpdateMetadata($contentId: String!, $metadata: JSON!) {
-    contentUpdateMetadata(contentId: $contentId, metadata: $metadata) {
+export interface ContentUpdateMutationResult {
```
Ostrzy: Isn't this duplicated? It looks the same as `ContentUpdateMetadataMutationResponse`? Also it looks out of place in the file, it should be below mutation, no? 🤔

packages/unique-api/src/ingestion/ingestion.queries.ts
```diff
@@ -79,29 +79,38 @@ export const CONTENT_UPSERT_MUTATION = gql`
...
+    contentUpdate(contentId: $contentId, input: { expiresAt: $expiresAt, metadata: $metadata }) {
       id
       metadata
       ingestionState
     }
   }
 `;
 
-export interface ContentUpdateMetadataMutationInput {
+export interface ContentUpdateInput {
```
Ostrzy: This looks inconsistent, no? 🤔

services/outlook-semantic-mcp/src/utils/date/get-message-expiration-date.ts
```diff
@@ -0,0 +1,14 @@
...
+  retentionWindowInDays,
+}: {
+  receivedDateTime: string | Date;
+  retentionWindowInDays: number;
+}): Date => {
+  // End of utc day.
+  const currentReceivedAt = new Date(receivedDateTime);
+  currentReceivedAt.setUTCHours(0, 0, 0, 0);
+  currentReceivedAt.setUTCDate(currentReceivedAt.getUTCDate() + retentionWindowInDays + 1);
+  currentReceivedAt.setUTCMilliseconds(-1);
```
Ostrzy: Would be nice to comment on why we're doing that like this :)

Ostrzy (approved):

Ostrzy:
Still some legit comments by CC 🙂 

## Branch Review: Content Expiration (`UN-16564`)                                                                                                                             
                                                                                                                                                                              
  ### Critical                                                                                                                                                                  
   
  #### 1. `subDays` uses local-time methods instead of UTC                                                                                                                      
  **`services/outlook-semantic-mcp/src/utils/date/sub-days.ts`**                                                                                                              
                                                                                                                                                                                
  `date.setDate(date.getDate() - days)` operates in the server's local timezone. Every sibling function uses explicit UTC methods. If the server runs in a non-UTC timezone near
   midnight, the subtraction can land on the wrong calendar day.                                                                                                                
                                                                                                                                                                                
  **Fix:** `date.setUTCDate(date.getUTCDate() - days)`                                                                                                                        

  ---

  #### 2. `update()` passes `Date` to GraphQL without `.toISOString()`                                                                                                          
  **`packages/unique-api/src/ingestion/ingestion.service.ts:35-43`**
                                                                                                                                                                                
  `registerContent` and `finalizeIngestion` both explicitly call `expiresAt?.toISOString()` before sending to GraphQL. The new `update()` method passes the raw `Date` object,  
  relying on implicit `JSON.stringify` serialization. Inconsistent and fragile.                                                                                                 
                                                                                                                                                                                
  ---                                                                                                                                                                         

  #### 3. Retention check inside fail-open `try/catch`
  **`services/outlook-semantic-mcp/src/features/inbox/process-email/utils/should-skip-email.ts:28-53`**
                                                                                                                                                                                
  `shouldSkipEmail` wraps everything in a `try/catch` returning `{ skip: false }`. If `computeRetentionCutoffDate` throws (e.g. `NaN` days), the email silently passes the      
  retention filter. Zod validation mitigates this at config-parse time, but fail-open around a data retention boundary is risky.                                                
                                                                                                                                                                                
  ---                                                                                                                                                                         

  ### Suggestions

  #### 4. `computeRetentionCutoffDate` is impure — recomputed per email
  **`services/outlook-semantic-mcp/src/utils/date/compute-retention-cutoff-date.ts`**
                                                                                                                                                                                
  Calls `new Date()` internally on every invocation. Within a batch spanning UTC midnight, different emails get different cutoffs. Consider computing once per batch or         
  accepting a `now` parameter.                                                                                                                                                  
                                                                                                                                                                                
  ---                                                                                                                                                                         

  #### 5. `getExpirationDate` — misleading variable name
  **`services/outlook-semantic-mcp/src/utils/date/get-message-expiration-date.ts:9`**
                                                                                                                                                                                
  `currentReceivedAt` is mutated into an expiration date. The `setUTCMilliseconds(-1)` trick is non-obvious. Consider renaming the variable and reusing `startOfUtcDay`.        
                                                                                                                                                                                
  ---                                                                                                                                                                           
                                                                                                                                                                              
  #### 6. Documentation mismatch
  **`services/outlook-semantic-mcp/docs/operator/configuration.md:294`**
                                                                                                                                                                                
  Prose says *"The example uses 180 days"* but the YAML example uses `retentionWindowInDays: 95`.                                                                               
                                                                                                                                                                                
  ---                                                                                                                                                                           
                                                                                                                                                                              
  #### 7. Stale type alias
  **`packages/unique-api/src/ingestion/ingestion.service.ts:9`**

  `ContentUpdateMetadataMutationResponse` is re-exported as `ContentUpdateMutationResponse` instead of using the new `ContentUpdateMutationResult` directly. Incomplete rename. 
   
  ---                                                                                                                                                                           
                                                                                                                                                                              
  #### 8. Mixed function declaration styles in `utils/date/`
  `computeRetentionCutoffDate` uses `export function`, the other three use `export const =>`. Within a single new module, pick one.
                                                                                                                                                                                
  ---
                                                                                                                                                                                
  #### 9. String-based date comparison is fragile                                                                                                                             
  **`services/outlook-semantic-mcp/src/features/process-email/process-email.command.ts:171`**
                                                                                                                                                                                
  Compares `expiresAt` via ISO string equality. Millisecond precision differences could cause false mismatches. `.getTime()` would be more robust.

packages/unique-api/src/ingestion/ingestion.queries.ts
nicolae536: Fixed

packages/unique-api/src/ingestion/ingestion.queries.ts
nicolae536: Fixed

services/outlook-semantic-mcp/src/utils/date/get-message-expiration-date.ts
nicolae536: Comment added.

packages/unique-api/src/ingestion/ingestion.queries.ts
```diff
@@ -78,29 +79,30 @@ export const CONTENT_UPSERT_MUTATION = gql`
...
 
-export interface ContentUpdateMetadataMutationInput {
+export interface ContentUpdateMutationInput {
   contentId: string;
-  metadata: Record<string, ContentMetadataValue>;
+  metadata?: Record<string, ContentMetadataValue>;
+  expiresAt?: Date;
 }
 
-export interface ContentUpdateMetadataResponse {
```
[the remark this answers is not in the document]
nicolae536: Fixed

services/outlook-semantic-mcp/src/features/process-email/process-email.command.ts
```diff
@@ -163,21 +164,31 @@ export class ProcessEmailCommand {
...
-      if (metadata.lastModifiedDateTime === file.metadata?.lastModifiedDateTime) {
+      const expirationDate = getMessageExpirationDate({
+        receivedDateTime: graphMessage.receivedDateTime,
+        retentionWindowInDays: filters.retentionWindowInDays,
+      });
+      const fileExpirationTimeStamp = file.expiresAt ? new Date(file.expiresAt).getTime() : -1;
+
+      if (
+        metadata.lastModifiedDateTime === file.metadata?.lastModifiedDateTime &&
+        expirationDate.getTime() === fileExpirationTimeStamp
```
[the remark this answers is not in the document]
nicolae536: Fixed

nicolae536:
3 => I'm not sure what he want's if there is no date zod will die before reaching this call so we are guarding against a non existent case
4 => I'm not sure if we need to do something there cause if a batch spans over a long time what's the point in ingesting something which is already expired ?

PR #435, opened 2026-04-06
feat(confluence-connector): set external ID on root scope and validate instance ownership

lorand93:
## Summary

Prevents misconfiguration where the same Confluence instance is connected to multiple Unique orgs, or the same root scope is pointed at a different instance.

On the first sync cycle the connector resolves a stable instance identifier and claims the root scope by setting its `externalId`. On subsequent syncs it validates the claim matches. A mismatch is a fatal error that stops the sync.

- Cloud uses `cloudId` from tenant config
- Data Center fetches the instance UUID from `GET /rest/applinks/1.0/manifest` (no auth required)

Closes UN-18352

## Ownership validation flow

```mermaid
flowchart TD
    A[initialize] --> B[Fetch root scope]
    B --> C[Resolve instance identifier]
    C --> D{rootScope.externalId?}
    D -- "null" --> E[Initial sync: claim ownership]
    E --> F[updateExternalId]
    F --> G[Continue sync]
    D -- "matches expected" --> G
    D -- "differs" --> H[Fatal error: ownership mismatch]
```

## Instance identifier resolution

```mermaid
flowchart LR
    subgraph Cloud
        CC[CloudConfluenceApiClient] --> CR["Return { type: 'cloud', id: config.cloudId }"]
    end
    subgraph Data Center
        DC[DataCenterConfluenceApiClient] --> DR["GET /rest/applinks/1.0/manifest"]
        DR --> DP["Return { type: 'data-center', id: manifest.id }"]
    end

    CR --> EXT["buildRootScopeExternalId()"]
    DP --> EXT
    EXT --> FMT["confc:cloud:uuid or confc:dc:uuid"]
```

## External ID format

| Instance type | Source | Format | Example |
|---|---|---|---|
| Cloud | `cloudId` from config | `confc:cloud:<id>` | `confc:cloud:a1b2c3d4-...` |
| Data Center | `/rest/applinks/1.0/manifest` | `confc:dc:<id>` | `confc:dc:92b23664-...` |

## Test plan
- [x] Cloud returns identifier from config without HTTP call (2 tests)
- [x] DC calls manifest endpoint, validates response (7 tests)
- [x] `buildRootScopeExternalId` produces correct format (2 tests)
- [x] Initial sync claims root scope ownership (2 tests)
- [x] Matching externalId proceeds normally (1 test)
- [x] Mismatched externalId throws fatal error (1 test)
- [x] Claim failure is non-fatal (1 test)
- [x] Instance identifier is cached across sync cycles (1 test)
- [x] All 273 confluence-connector tests pass
- [x] biome check clean, tsc clean


lorand93:
Missing sanitizeError -> created a separate ticket for this to do the same  as sharepoint-connector https://unique-ch.atlassian.net/browse/UN-19205
 Root scope migration -> we have a ticket for this  https://unique-ch.atlassian.net/browse/UN-18993
Fail on cross-org duplicate -> done: validateOwnership now throws on claim failure (line 107) with comment explaining why

image

services/confluence-connector/src/synchronization/scope-management.service.ts
```diff
@@ -46,7 +73,47 @@ export class ScopeManagementService {
...
+
+    if (!rootScope.externalId) {
+      // Claim fails fatally: if updateExternalId rejects (e.g. the externalId is already
+      // taken within this org), the sync must not proceed to avoid data conflicts.
+      try {
+        const updatedScope = await this.uniqueApiClient.scopes.updateExternalId(
+          rootScope.id,
+          expectedExternalId,
+        );
+        rootScope.externalId = updatedScope.externalId;
```
Ostrzy: Maybe it would make sense to add here as well the 
```ts
    assert.strictEqual(updatedScope.externalId, expectedExternalId, `Root scope ownership mismatch`);
```

services/confluence-connector/src/synchronization/scope-management.service.ts
```diff
@@ -46,7 +73,47 @@ export class ScopeManagementService {
 
     const rootScopePath = `/${pathSegments.join('/')}`;
     this.logger.log({ rootScopePath, msg: 'Scope management initialized' });
-    return rootScopePath;
+    return { rootScopePath, isInitialSync };
+  }
+
+  private async validateOwnership(rootScope: {
```
Ostrzy, line 79: Why does `validateOwnership` return `boolean` indicating whether it's the initial sync? 😅 It's, to put it lightly, weird 😛

services/confluence-connector/src/synchronization/scope-management.service.ts
```diff
@@ -46,7 +73,47 @@ export class ScopeManagementService {
 
     const rootScopePath = `/${pathSegments.join('/')}`;
     this.logger.log({ rootScopePath, msg: 'Scope management initialized' });
-    return rootScopePath;
+    return { rootScopePath, isInitialSync };
```
Ostrzy: Can we just drop the `isInitialSync`? We don't use it, don't care and can easily add it if it's needed 🙂 YAGNI!

Ostrzy (approved):
Few small comments that I think would be nice to address 🙂

services/confluence-connector/src/synchronization/scope-management.service.ts
```diff
@@ -46,7 +73,47 @@ export class ScopeManagementService {
...
+  private async validateOwnership(rootScope: {
+    id: string;
+    externalId: string | null;
+  }): Promise<boolean> {
+    const instanceId = await this.getInstanceIdentifier();
+    const expectedExternalId = buildRootScopeExternalId(instanceId.type, instanceId.id);
+
+    if (!rootScope.externalId) {
+      // Claim fails fatally: if updateExternalId rejects (e.g. the externalId is already
+      // taken within this org), the sync must not proceed to avoid data conflicts.
```
Ostrzy, line 88: Wait, but if scope externalID is per org, it means we will not fail when we would sync the same confluence instance into two separate orgs, no? :thinking_face:

Ostrzy (changes requested):

services/confluence-connector/src/synchronization/scope-management.service.ts
lorand93: dropped it, thought to keep it as spc has it, but we have no use for it yet here so i get your point

services/confluence-connector/src/synchronization/scope-management.service.ts
lorand93: it's a good point, I added validateUniqueConfluenceInstances and now we shouldn't get the same confluence instance configured in any two tenants

Ostrzy (approved):

PR #459, opened 2026-04-09
feat(sharepoint-connector,deps): add /health endpoint with sync, connectivity, and Unique API indicators

Ostrzy:
## Summary
- Adds a `GET /health` endpoint using `@nestjs/terminus` with three health indicators:
  - **Sync** — tracks a sliding window of recent sync results and reports down when per-site failure rate exceeds a configurable threshold
  - **Connectivity** — pings Microsoft Graph and SharePoint without auth to verify network reachability (any HTTP response = reachable)
  - **Unique API** — pings ingestion and scope-management GraphQL endpoints with auth, reports specific HTTP error codes on failure
- Introduces `SyncStatusStore` ring buffer to record sync outcomes, written by the scheduler after each run
- Returns per-site sync results from `synchronize()` so the health module can attribute failures to individual sites

Ref: UN-15710

## Test plan
- [x] Verify `GET /health` returns structured response with all three indicators
- [x] Turn off WiFi — connectivity indicator should report `down` with specific error codes (e.g. `ENOTFOUND`)
- [x] Revoke/break Zitadel credentials — uniqueApi indicator should report `AUTH_FAILURE`
- [x] Trigger sync failures — sync indicator should report `down` when failure threshold is exceeded
- [x] Run `pnpm test` for sharepoint-connector — all tests pass


services/sharepoint-connector/src/health/sync-health.indicator.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+    if (key === 'health.syncHistorySize') {
+      return HISTORY_SIZE;
+    }
+    if (key === 'health.syncSiteFailureThreshold') {
+      return THRESHOLD;
+    }
+    return undefined;
+  },
+} as unknown as ConfigService<Config, true>;
+
```
vabatta: `as unknown as ...` here and for the whole other batch below should instead use `TestSuites` (that we have) for proper testing; leave as-is, up to you if you want to fix, it's not a blocker.

vabatta (approved):
Overall good. I would have _nit picks_ for code writing, but in this AI era doesn't really matter so I skipped them.

Only one nit as per convention/tooling we use, up to you.

PR #488, opened 2026-04-22
fix(confluence-connector,unique-api,utils,logger,deps): sanitize errors in logging

lorand93:
closes https://unique-ch.atlassian.net/browse/UN-19205

## What

Error logs could leak tokens, content keys, and internal URLs. When a GraphQL call failed, the `graphql-request` `ClientError` embedded the full request (query + variables) into `error.message` and `error.stack`, and pino logged that straight through.

## How

- Added `sanitizeError` / `normalizeError` to `@unique-ag/utils`, ported from the existing SharePoint implementation. SharePoint is left untouched.
- Wrapped every raw-error log in the Confluence connector with `sanitizeError(...)`.
- Fixed the same leak upstream in `packages/unique-api` (graphql client, http client, files service, auth) and in `packages/utils` (`processInBatches`), since those logs fire before the consumer can sanitize.
- Log key stays `err:` so existing Loki queries and dashboards keep working.

## Test plan

- [x] type-check, lint, and tests green in `packages/utils`, `packages/unique-api`, and `services/confluence-connector`
- [x] no raw `err: error` / `error: error` left in touched source trees


lorand93:
bugbot review

packages/utils/src/normalize-error.ts
```diff
@@ -0,0 +1,62 @@
+import { ClientError } from 'graphql-request';
+import { serializeError } from 'serialize-error-cjs';
```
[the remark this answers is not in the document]
lorand93: thank god for bugbot, i did not think of this...

lorand93:
bugbot review

services/confluence-connector/src/synchronization/confluence-content-fetcher.ts
```diff
@@ -19,7 +20,7 @@ export class ConfluenceContentFetcher {
       this.logger.error({
         pageId: page.id,
         title: page.title,
-        err: error,
+        err: sanitizeError(error),
```
nicolae536: I don't think we want to change every log which contains error -> put it in the pino logger so that it's applied everywhere.

packages/unique-api/src/auth/unique-auth.ts
```diff
@@ -100,7 +101,7 @@ export class UniqueAuth implements UniqueAuthFacade {
     } catch (error) {
       this.logger.error({
         msg: 'Failed to acquire Unique API token from Zitadel',
-        error,
+        error: sanitizeError(error),
```
nicolae536: wrong key it should be `err`

packages/unique-api/src/clients/unique-graphql.client.ts
```diff
@@ -166,7 +167,10 @@ export class UniqueGraphqlClient {
         });
       }
 
-      this.logger.error({ msg: `Failed ${this.config.target} request (${operationName})`, error });
+      this.logger.error({
+        msg: `Failed ${this.config.target} request (${operationName})`,
+        error: sanitizeError(error),
```
nicolae536: wrong key it should be err

packages/unique-api/src/clients/unique-http.client.ts
```diff
@@ -162,7 +163,7 @@ export class UniqueHttpClient {
         });
       }
 
-      this.logger.error({ msg: 'Failed ingestion HTTP request', error });
+      this.logger.error({ msg: 'Failed ingestion HTTP request', error: sanitizeError(error) });
```
nicolae536: wrong key it should be `err`

packages/unique-api/src/files/files.service.ts
```diff
@@ -272,7 +272,7 @@ export class FilesService implements UniqueFilesFacade {
               msg: `${logPrefix} Failed to add single file access`,
               scopeId,
               permission,
-              error: singleError,
+              error: sanitizeError(singleError),
```
nicolae536: wrong key it should be `err`

packages/unique-api/src/files/files.service.ts
```diff
@@ -330,7 +330,7 @@ export class FilesService implements UniqueFilesFacade {
               msg: `${logPrefix} Failed to remove single file access`,
               scopeId,
               permission,
-              error: singleError,
+              error: sanitizeError(singleError),
```
nicolae536: wrong key it should be `err`

packages/utils/src/normalize-error.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Serialises an error for structured logging.
+ *
+ * For `graphql-request` `ClientError` instances the full request payload
+ * (query + raw variables) that the library embeds in `message` / `stack` is
+ * stripped and replaced with structured `graphqlErrors` and `statusCode`
+ * fields, so unsanitised variables never reach the logs.
+ */
+export function sanitizeError(error: unknown): object {
```
nicolae536: can we returns a proper type ? 🤔

packages/utils/src/normalize-error.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (error instanceof ClientError) {
+    const jsonDumpStart = error.message.indexOf(': {"response":');
+    const baseMessage = jsonDumpStart >= 0 ? error.message.slice(0, jsonDumpStart) : error.message;
```
nicolae536: Hmm while this fixes the current issue we should probably create a ticket or update this ticket -> https://unique-ch.atlassian.net/browse/UN-19018?search_id=908d2fd8-e77b-4c45-9e05-3438277406a7 basically we need a solution in general for sensitive details in the console - probably added to `pino` which should strip out any sensitive content.

packages/utils/src/process-in-batches.ts
```diff
@@ -44,7 +45,7 @@ export async function processInBatches<TInput, TOutput>({
       logger.error({
         batchIndex: index,
         batchSize: batch.length,
-        err: error,
+        err: sanitizeError(error),
```
nicolae536: I'm pretty sure at this point that we do not want to call this function is all the code base

nicolae536 (changes requested):

packages/unique-api/src/auth/unique-auth.ts
lorand93: i figured I shouldn't change the logging key to err in this PR. but since you flagged it I will change it :)

packages/utils/src/normalize-error.ts
lorand93: Agree. I’ll keep this PR scoped to the current GraphQL `ClientError` leak

Moving this into shared logger affects other connector log shapes, especially SharePoint

lorand93:
bugbot review

packages/utils/src/normalize-error.ts
lorand93: added SanitizedError

lorand93:
bugbot review

packages/utils/src/normalize-error.ts
```diff
@@ -0,0 +1,77 @@
...
+        message: e.message,
+        path: e.path,
+        code: e.extensions?.code,
+      })),
+      statusCode: error.response.status,
+    };
+  }
+
+  return serializeError(normalizeError(error));
+}
```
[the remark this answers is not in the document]
lorand93: yes, I know about it. I removed the logger-local sanitizer and wired logger to use the shared @unique-ag/utils implementation. The SharePoint copy predates this PR, so I’m leaving that out of scope here.

packages/logger/src/options.ts
```diff
@@ -1,7 +1,9 @@
 import { IncomingMessage } from 'node:http';
 import path from 'node:path';
+import { sanitizeError } from '@unique-ag/utils';
```
nicolae536: Shouldn't we move this in the `logger` package and keep it internal ? to avoid the gotcha that llm generates code which contains this sanitizeError ?

packages/logger/src/options.ts
```diff
@@ -27,10 +29,14 @@ export const defaultLoggerOptions: Params = {
       ],
       censor: () => '[Redacted]',
     },
+    serializers: {
+      err: sanitizePinoError,
```
nicolae536, line 33: just as an idea -> you can create a ticket for it and later on we fix it. wouldn't it be normal that we pass to the logger module the secrets ? And then on every logged message we do
```
for (const secret of secrets) {
   message = message.replace(secret, `[Redacted]`);
}
```
like this the secrets are 100% out of the logged response - we can create a ticket fix it like this.

packages/utils/src/normalize-error.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  if (typeof error === 'symbol') {
+    return new Error(error.toString());
+  }
+  if (typeof error === 'function') {
+    return new Error(error.toString());
+  }
+
+  try {
+    return new Error(JSON.stringify(error));
```
nicolae536, line 25: I'm just curious why we want to handle serialization - instead of doing `clone(error)` ? 🤔 doesn't pino handle this if the object is in `cause`  of error ?

packages/logger/src/options.ts
lorand93: Done. moved sanitizeError into the logger package as an internal implementation detail and removed it from @unique-ag/utils exports. It's no longer part of the public API.

packages/utils/src/normalize-error.ts
lorand93: think It would be possible to do what claude recommends here:
-the one optional improvement Opus suggests - passing the original as cause alongside the JSON.stringify message:

  return new Error(JSON.stringify(error), { cause: error });
  
  wdyt?

packages/logger/src/options.ts
lorand93: it's a good idea.
https://unique-ch.atlassian.net/browse/UN-20407

packages/utils/src/normalize-error.ts
```diff
@@ -0,0 +1,30 @@
+export function normalizeError(error: unknown): Error {
+  if (error instanceof Error) {
+    return error;
+  }
```
nicolae536, line 4: Shouldn't this be inside the `logger` package and the package should not export it

packages/utils/src/normalize-error.ts
nicolae536: you can try that -> cause serializing them is usualy hard we can try the following options
- Using cause as claude suggested
- look for a library which can handle `JSON.stringify` even if the object has circular references I remember there are some options in `npm` but I would keep the `try {} catch {}` anyway because of paranoia if we go this route

nicolae536 (approved):

packages/utils/src/normalize-error.ts
lorand93: good catch!

packages/utils/src/normalize-error.ts
lorand93: done 👍  thanks for the input

lorand93:
oupsy, didn't push the changes before hitting merge...

PR #494, opened 2026-04-28
feat(confluence-connector,unique-api): ingest images and switch attachment filter to MIME types

lorand93:
## Summary

Resolves [UN-19064](https://unique-ch.atlassian.net/browse/UN-19064). Enables ingestion of images embedded in Confluence pages (PNG/JPEG) and switches attachment filtering from extension-based to MIME-type-based, resolving the existing TODO in `confluence-page-scanner.ts`.

Embedded images in Confluence are already regular page attachments in the data model (the body just references them via `<ac:image><ri:attachment/></ac:image>`). The connector's attachment pipeline was already wired to discover, stream, and register them with the real `mediaType`. The only thing blocking image ingestion was the extension allowlist, which has now been replaced by a MIME-type allowlist aligned with `node-ingestion`'s `allowedMimeTypesMap`.

- **`png`, `jpg`, `jpeg` are now ingested by default** when attachment ingestion is enabled.
- **Filter switched from filename extension to `attachment.extensions.mediaType`**, with charset-suffix stripping (`image/png; charset=binary`) and whitespace trimming.
- Removed the `extractExtension` helper and the `// TODO switch to mime type` comment.
- Docs updated across `README.md`, `operator/configuration.md`, `operator/README.md`, `technical/README.md`, `technical/flows.md`, and `faq.md`.

## Notes for operators

For images to produce searchable chunks, the destination scope's `ingestionConfig.jpgReadMode` must be set to `DOC_INTELLIGENCE_DEFAULT`. Without it, image uploads succeed but no chunks are produced.

Other image formats (GIF, WebP, SVG, HEIC, BMP, TIFF) are not currently supported by the Unique ingestion service and should be left out of `allowedMimeTypes`.

## Test plan

- [x] `pnpm exec tsc --noEmit` clean
- [x] `pnpm exec biome check src` clean
- [x] `pnpm exec vitest run` — 346 tests passing (3 new edge-case tests added: charset suffix, whitespace, octet-stream rejection)
- [ ] Smoke test against a real Confluence page with an embedded PNG: attachment discovered, uploaded with `image/png`, chunks produced (requires scope with `jpgReadMode = DOC_INTELLIGENCE_DEFAULT`)
- [ ] Verify behavior of a tenant YAML still using `allowedExtensions:` (silent fallback to defaults — flagged below)

BREAKING CHANGE: `ingestion.attachments.allowedExtensions` has been renamed to `ingestion.attachments.allowedMimeTypes` and now expects MIME type strings instead of file extensions. Operators with custom values in their tenant YAML must migrate. Default values are now MIME types covering PDF, DOCX, XLSX, PPTX, plain text, CSV, HTML, PNG, and JPEG. Legacy `.ppt` (Office 97-2003, `application/vnd.ms-powerpoint`) is no longer in defaults; it was previously matched by extension but rejected by `node-ingestion`'s public API gate, so it was already non-functional end-to-end. The connector is still alpha (`prerelease: true`), so no compatibility shim is provided.

[UN-19064]: https://unique-ch.atlassian.net/browse/UN-19064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lorand93:
bugbot review

services/confluence-connector/src/synchronization/__tests__/confluence-page-scanner.spec.ts
```diff
@@ -587,6 +645,33 @@ describe('ConfluencePageScanner', () => {
       expect(result.attachments.map((a) => a.id)).toEqual(['att-1']);
     });
 
+    it('rejects application/octet-stream attachments by default', async () => {
```
nicolae536: shouldn't be - skips attachemnts...  ? I would expect an error

services/confluence-connector/docs/technical/flows.md
```diff
@@ -192,10 +192,14 @@ Pages labeled with `ingestAllLabel` trigger a descendant search:
...
 
 An attachment is accepted if:
-- Its file extension is in the `allowedExtensions` list (default: `pdf`, `docx`, `xlsx`, `ppt`, `pptx`, `txt`, `csv`, `html`)
+- Its `mediaType` (reported by the Confluence API) is in the `allowedMimeTypes` list (defaults cover PDF, the major Office formats, plain text, CSV, HTML, PNG, and JPEG)
 - Its file size does not exceed `maxFileSizeMb` (default: 200 MB)
 - The `maxItemsToScan` capacity has not been exhausted (pages count first, attachments use remaining capacity)
 
+Images embedded in a page body (via the editor's drag/drop, paste, or "Insert image" actions) are stored by Confluence as regular page attachments and surface in the same `expand=children.attachment` results, so they are ingested through this path when `image/png` or `image/jpeg` is in `allowedMimeTypes`. Images inserted as external URLs are not attachments and are not ingested.
+
```
nicolae536, line 200: Just a question here -> are confluence scopes manageable by external users or just your service user can manage them -> I feel this should be put on the scopes as ingestion config instead of forcing it on each upload, I'm not really sure which version is better.

nicolae536 (approved):
Some small comments

services/confluence-connector/docs/technical/flows.md
lorand93: good point to raise, I also gave it a thought if I should add this in the step where we also add the external id on the root scope. 

I tested it with setting this on the scope level and then not passing it when we ingest and it didn't work. file  cont_y4iyuxrzthuoni8csgb740vo from scope         "scopeId": "scope_hcppricirgfyz05qliz6yfrh",

Also doing it this way gives up the option to send this on different mime types which is better than we have for setting it on scope level.

services/confluence-connector/src/synchronization/__tests__/confluence-page-scanner.spec.ts
lorand93: updated name

PR #505, opened 2026-05-04
fix(sharepoint-connector): log partial failures when deleting stale scopes

lorand93:
## Summary
- When a stale scope deletion completes but some folders fail to delete, surface those failures via a structured warning log so they're visible in observability tooling. Previously a partial failure was silently logged at debug level as a successful deletion.

## Test plan
- [ ] Verify warning log appears when `deleteScope` returns a non-empty `failedFolders` array
- [ ] Verify debug log still appears on full success

### DIFF
--- a/services/sharepoint-connector/src/sharepoint-synchronization/scope-management.service.ts
+++ b/services/sharepoint-connector/src/sharepoint-synchronization/scope-management.service.ts
@@ -245,8 +245,18 @@ export class ScopeManagementService {
 
     for (const scope of sortedStaleScopes) {
       try {
-        await this.uniqueScopesService.deleteScope(scope.id);
-        this.logger.debug(`${logPrefix} Deleted stale scope ${scope.id}`);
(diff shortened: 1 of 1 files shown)


florianrueegg (approved):

services/sharepoint-connector/src/sharepoint-synchronization/scope-management.service.ts
```diff
@@ -245,8 +245,18 @@ export class ScopeManagementService {
...
+          this.logger.warn({
+            msg: `${logPrefix} Failed to delete stale scope ${scope.id}`,
+            failedFolders: result.failedFolders.map((f) => ({
+              id: f.id,
+              reason: f.failReason,
+            })),
+          });
+        } else {
+          this.logger.debug(`${logPrefix} Deleted stale scope ${scope.id}`);
+        }
```
Ostrzy, line 257: I would ignore the debug here - if we don't log anything stuff clearly worked as expected 🙂

Ostrzy (approved):

PR #526, opened 2026-05-08
feat(sharepoint-connector): configurable site-sync defaults via Helm

Ostrzy:
## Summary

- Add a `sharepoint.siteDefaults` block (rendered by Helm) so deployment-level defaults fill in any per-site field except `siteId`.
- Per-site values from the SharePoint list or `config_file` always win when set; empty/whitespace cells fall through to deployment defaults, then to existing schema defaults.
- Backward compatible: tenants without `siteDefaults` keep today's behavior; failure mode (one bad row aborts the whole load) is preserved.
- Touches `sharepoint.schema.ts`, a new merger module, `SitesConfigurationService`, both example tenant configs, and the Helm chart `values.yaml` + a new helm-unittest case.

Ticket: UN-20464


services/sharepoint-connector/src/config/site-config-merger.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export function mergeSiteWithDefaults(
+  partialSite: PartialSiteConfig,
+  siteDefaults: SiteDefaults,
+  rowIdentifier: string,
+): SiteConfig {
+  const merged: Record<keyof SiteConfig, unknown> = {
+    siteId: partialSite.siteId,
+    syncColumnName: coalesce(partialSite.syncColumnName, siteDefaults.syncColumnName),
+    ingestionMode: coalesce(partialSite.ingestionMode, siteDefaults.ingestionMode),
```
nicolae536, line 47: I'm not sure I like the approach why not zod defaults ? 🤔 and make entire template values optional ? Anyway it's just a style thing so feel free to ignore my comment

nicolae536 (approved):

services/sharepoint-connector/src/config/site-config-merger.ts
Ostrzy: As we discussed, we would need to make dynamic schemas, because defaults are also in the tenant file. Merging them like this is cleaner I think 🙂

PR #527, opened 2026-05-08
test(confluence-connector): integration test suite for end-to-end sync scenarios

lorand93:
## Why
This PR adds end-to-end integration tests that drive `synchronize()` against realistic state on both sides, locking down **38 user-observable scenarios**.

## How

The real synchronization stack runs unmodified. The two HTTP boundaries we don't own are replaced with stateful in-memory fakes that implement the same TypeScript interface as the real clients. Blob uploads are intercepted with `undici.MockAgent`, so the real upload path runs.

A test reads:
1. Define an initial state (Confluence + Unique).
2. Call `runSync()` (or `runDelete()`).
3. Snapshot Unique's state and assert against it.

```mermaid
flowchart TB
    subgraph ctx["buildScenarioContext(scenario)"]
        direction TB
        subgraph services["Real services — wired exactly like TenantRegistry"]
            direction TB
            sync["ConfluenceSynchronizationService"]
            scanner["ConfluencePageScanner"]
            fetcher["ConfluenceContentFetcher"]
            diff["FileDiffService"]
            ingest["IngestionService"]
            scope["ScopeManagementService"]
            del["TenantDeleteService"]
            sync --> scanner
            sync --> fetcher
            sync --> diff
            sync --> ingest
            sync --> scope
        end

        confluence["FakeConfluenceApi<br/><i>extends real client</i><br/>+ mutation API"]
        unique["FakeUniqueApi<br/><i>implements UniqueApiClient</i><br/>+ per-space performFileDiff"]
        blob["FakeBlobStorage<br/><i>undici MockAgent</i><br/>intercepts PUTs against writeUrl"]

        scanner --> confluence
        fetcher --> confluence
        diff --> unique
        ingest --> unique
        scope --> unique
        del --> unique
        ingest -. "PUT writeUrl" .-> blob
        blob -. "stores bytes" .-> unique
    end

    classDef fake fill:#fff4e6,stroke:#d97706,color:#7c2d12
    classDef real fill:#eef2ff,stroke:#4338ca,color:#1e1b4b
    class confluence,unique,blob fake
    class sync,scanner,fetcher,diff,ingest,scope,del real
```

```mermaid
flowchart LR
    A["<b>Scenario</b><br/>defineScenario({...})<br/>aPage · aSpace · aPageFile · …"]
      --> B["<b>runSync()</b> / <b>runDelete()</b>"]
      --> C["<b>getUniqueState()</b><br/>sorted, diff-friendly view<br/>scopes by path · files by key"]
```

### Fakes, not mocks

`FakeConfluenceApi` extends the real Cloud client and overrides only the network seam, so the real CQL handling, descendant fan-out, and 25-attachment inline cap + v2 pagination all run. `FakeUniqueApi` implements `performFileDiff` with the same per-space, prefix-match semantics the real API uses. `FakeBlobStorage` runs real `fetch(writeUrl, { method: 'PUT' })` and lets `undici.MockAgent` intercept. The diff loop and the upload + finalize path are end-to-end-tested without HTTP. Per-item failures are injected with `vi.spyOn` on the real client methods rather than bespoke hooks.

Assertions go through two shared expecters, `expectIngested` / `expectNotIngested`, which take page and attachment keys and check (order-independently) that they did or did not reach Unique.

## Scenarios

| Spec | What it asserts |
|---|---|
| `single-page-sync` | A single `ai-ingest`-labeled page becomes one HTML file under a space scope. |
| `single-page-sync` | A labeled page + its PDF attachment ingest as two files; attachment bytes match exactly. |
| `single-page-sync` | Unlabeled siblings in the same space are not ingested. |
| `subtree-sync` | An `ai-ingest-all` root pulls in its entire descendant tree. |
| `subtree-sync` | Multiple `ai-ingest-all` roots in one space all traverse independently. |
| `subtree-sync` | Multiple spaces produce independent space scopes and partial-key namespaces. |
| `subtree-sync` | A 50-descendant tree fully ingests — no silent drop in the fan-out. |
| `subtree-sync` | `ai-ingest` and `ai-ingest-all` coexist in one space; their union is ingested with no duplicates. |
| `subtree-sync` | An `ai-ingest` page nested under an unlabeled (non-ingested) parent is still ingested. |
| `update-content` | A page version bump rewrites the existing file in place under the same key. |
| `update-content` | The `validateNoAccidentalFullDeletion` carve-out allows full content replacement when no submitted key overlaps a deleted key. |
| `delete-content` | A page removed from Confluence is deleted from Unique on the next sync. |
| `delete-content` | An attachment removed from a page is deleted while the page itself stays. |
| `delete-space` | A space scope and all its files are deleted when the space loses all labeled content. |
| `delete-space` | A space scope and all its files are deleted when the space itself is gone from Confluence. |
| `delete-space` | A space scope with current labeled content is preserved across syncs (no-op baseline). |
| `delete-tenant` | The tenant deletion flow removes every child scope and file but preserves the root scope. |
| `delete-tenant` | The root externalId is cleared only after a successful cleanup, used as the "done" marker. |
| `delete-tenant` | A run that fails partway keeps the externalId set; a successful retry then clears it. |
| `delete-tenant` | A second deletion is a no-op (returns `skipped` with `already_cleaned_up`). |
| `root-scope-ownership` | An unowned root scope is claimed on first sync (`externalId` written). |
| `root-scope-ownership` | An already-owned root proceeds without touching the externalId. |
| `root-scope-ownership` | A foreign-owned root aborts the sync and writes nothing. |
| `root-scope-migration` | When the previous root's externalId is already cleared there is nothing to migrate: the new root is claimed and synced; orphaned old-root content is left for the operator. |
| `root-scope-migration` | When the previous root still holds the externalId, the sync migrates its child scopes to the new root, deletes the old root, and claims the externalId. |
| `content-types` | Blog posts ingest the same as pages. |
| `content-types` | `database`, `whiteboard`, and `embed` content is skipped even when directly labeled. |
| `content-types` | An `ai-ingest-all` database/whiteboard/embed root has its descendants ingested even though the root itself is skipped. |
| `attachments` | PNG and JPEG attachments ingest as first-class files alongside the page. |
| `attachments` | The MIME allowlist excludes attachments outside it. |
| `attachments` | Attachments larger than `maxFileSizeMb` are filtered at scan time. |
| `attachments` | When `imageOcrEnabled` is on, image attachments are registered with `jpgReadMode: DOC_INTELLIGENCE_DEFAULT`; non-image attachments are unaffected. |
| `attachments` | A 50-attachment page fully ingests — exercises the 25-inline cap + v2 pagination, no silent drop. |
| `attachments` | When attachments are disabled, only the HTML page reaches Unique. |
| `failure-isolation` | One page failing to fetch from Confluence doesn't abort the sync. |
| `failure-isolation` | One Unique `registerContent` failure doesn't abort the sync. |
| `failure-isolation` | An attachment download failure after registration triggers `cleanupFailedRegistration`, leaving no orphan. |
| `cross-space-move` | A page relocated to a different space is deleted from its old space and re-ingested under the new one. Confluence keys are space-scoped, so there is no single "move" — the diff sees new-here plus deleted-there. |

## Unit tests

Alongside the integration suite, this PR also fills unit-test gaps for the config schemas (`confluence`, `ingestion`, `processing`), `MetricsService`, `ProxyService`, paginated Confluence fetch, `http-util`, and `Redacted`.

## Review notes

- **No false positives.** Each test asserts user-observable state (keys, body bytes, mimetype, scope tree, externalId, ingestionConfig, sync status). A regression in the path under test would visibly fail.
- **No fictional behavior.** The fakes model only what the real APIs do. Notably, Confluence's content keys are space-scoped and its file diff runs per space, so there is no cross-key "move" — a relocated page is a delete + reingest (see `cross-space-move`). The `movedFiles` diff category never fires for Confluence.
- **One documented duplicate.** Two `delete-space` tests share the same `cleanupRemovedSpaces` code path because the connector cannot distinguish "no labeled content" from "space gone" at sync time. Both kept so each operator-facing scenario appears by name. Comment explains why.
- **One dismissed gap.** `validateNoAccidentalFullDeletion`'s abort branches are unreachable through `synchronize()` (the per-space diff loop only iterates spaces with submitted items, so a full deletion always carries new files; the key-overlap branch is a defensive guard against an internal bug). Already covered by `file-diff.service.spec.ts`.

## Test plan

- [x] `pnpm test:integration` — 38/38 pass (~1.3s)
- [x] `pnpm test` — 483/483 pass
- [x] `pnpm check-all` — types + biome clean
- [ ] CI green


lorand93:
bugbot review

services/confluence-connector/test/integration/__tests__/subtree-sync.integration-spec.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  // The canonical subtree-sync flow: label the root, get the whole tree.
+  it('ingests a labeled root page together with its entire descendant tree', async () => {
+    const scenario = defineScenario({
+      confluence: {
+        spaces: [aSpace()],
+        pages: [
+          aPage({ id: 'root', title: 'Handbook', labels: ['ai-ingest-all'] }),
+          aPage({ id: 'child-a', parentId: 'root', title: 'Onboarding' }),
+          aPage({ id: 'child-b', parentId: 'root', title: 'Benefits' }),
```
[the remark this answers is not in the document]
lorand93: fixed

services/confluence-connector/test/integration/scenario/builders.ts
```diff
@@ -0,0 +1,161 @@
...
+ * Entity builders that produce well-formed scenario primitives with sensible defaults.
+ *
+ * The override parameter lets tests focus on what matters:
+ *
+ *     aPage({ id: 'p1', body: '<p>Custom</p>' })
+ *
+ * instead of repeating spaceKey, labels, versionWhen, etc. on every page.
+ */
+
+export function aSpace(overrides: Partial<ScenarioSpace> = {}): ScenarioSpace {
```
Ostrzy: Lets just make these `space`, `page` etc, without that `a`

services/confluence-connector/test/integration/scenario/builders.ts
```diff
@@ -0,0 +1,161 @@
...
+    ...overrides,
+  };
+}
+
+/**
+ * Domain-flavored shorthand for seeding a space scope under the configured root.
+ * Produces the proper externalId (`confc:<tenant>:<spaceId>:<spaceKey>`) so that
+ * `cleanupRemovedSpaces` can identify it the same way it does in production.
+ */
+export function aSpaceScope(opts: {
```
Ostrzy: Let's split these builders into confluence builders and unique builders just so we're clear

services/confluence-connector/test/integration/scenario/scenario.types.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export interface Scenario {
+  tenant: ScenarioTenantConfig;
+  confluence: ScenarioConfluence;
+  unique: ScenarioUnique;
+}
+
+/** Resolved tenant config in the same shape used by the production services. */
+export type ResolvedTenantConfig = TenantConfig;
```
Ostrzy: It seems to not be used

services/confluence-connector/test/integration/scenario-context/scenario-context.ts
```diff
@@ -0,0 +1,136 @@
...
+  readonly tenant: TenantContext;
+  readonly confluence: FakeConfluenceApi;
+  readonly unique: FakeUniqueApi;
+  readonly metrics: Metrics;
+  runSync(): Promise<SyncResult>;
+  /**
+   * Run the tenant-deletion flow (used when a tenant's status flips to
+   * `deleted`). Wipes child scopes and content while preserving the root scope.
+   */
+  runDelete(): Promise<DeleteResult>;
```
Ostrzy, line 26: Please check if we need it, I would expect that deletion happens as part of `runSync` as it is with the standard flow

services/confluence-connector/test/integration/fakes/fake-blob-storage.ts
```diff
@@ -0,0 +1,83 @@
...
+      })
+      .persist();
+  }
+
+  public asDispatcher(): Dispatcher {
+    return this.mockAgent;
+  }
+
+  public async close(): Promise<void> {
+    await this.mockAgent.close();
```
Ostrzy: Double check if we need to run this close or maybe not

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+        byteSize: file.byteSize,
+        mimeType: file.mimeType,
+        ownerType: 'SCOPE',
+        ownerId: file.scopeId ?? 'unknown',
+        fileAccess: [],
+        expiresAt: null,
+        ingestionState: IngestionState.Finished,
+        metadata: file.metadata ?? null,
+        body: file.body,
+        updatedAt: file.updatedAt ?? new Date(0).toISOString(),
```
Ostrzy: Maybe worth a comment why we have this `new Date(0)` cause it's not obvious

services/confluence-connector/test/integration/scenario-context/scenario-context.ts
```diff
@@ -0,0 +1,136 @@
...
+    runSync: () => tenantStorage.run(tenant, () => syncService.synchronize()),
+    runDelete: () => tenantStorage.run(tenant, () => deleteService.deleteTenantContent()),
+    close: () => blobStorage.close(),
+  };
+}
+
+/**
+ * ScopeManagementService.initialize() expects a pre-existing root scope at
+ * `ingestion.scopeId`. If the scenario didn't seed it, create it here so tests
+ * don't need to repeat that boilerplate.
```
Ostrzy: I think we should either crash it in define scenario or default it so we don't need this weird-ish case

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+    this.failures.registerContent.set(key, error);
+  }
+
+  public failOnFinalizeIngestion(key: string, error: Error): void {
+    this.failures.finalizeIngestion.set(key, error);
+  }
+
+  public failOnPerformFileDiff(error: Error): void {
+    this.failures.performFileDiff = error;
+  }
```
Ostrzy: We can add them if we have the future scenarios

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+        return { id: scope.id, externalId: scope.externalId };
+      },
+      updateParent: async (scopeId, newParentId) => {
+        const scope = this.requireScope(scopeId);
+        scope.parentId = newParentId;
+        return { id: scope.id, parentId: scope.parentId };
+      },
+      listChildren: async (parentId) =>
+        [...this.scopesById.values()].filter((scope) => scope.parentId === parentId),
+      createAccesses: async (_scopeId: string, _accesses: ScopeAccess[]) => {
```
Ostrzy, line 171: We should use the `notImplemented` consistently for all methods we do not use

services/confluence-connector/test/integration/scenario-context/unique-state.ts
```diff
@@ -0,0 +1,106 @@
...
+  const scopePathById = buildScopePathIndex(scopes);
+
+  const scopeStates: UniqueScopeState[] = scopes
+    .map((scope) => ({
+      id: scope.id,
+      name: scope.name,
+      path: scopePathById.get(scope.id) ?? scope.name,
+      externalId: scope.externalId,
+    }))
+    .sort((a, b) => a.path.localeCompare(b.path));
```
Ostrzy: `sortBy`

services/confluence-connector/test/integration/scenario-context/unique-state.ts
```diff
@@ -0,0 +1,106 @@
...
+      key: file.key,
+      byteSize: file.byteSize,
+      mimeType: file.mimeType,
+      metadata: file.metadata,
+      bodyHash: file.body ? sha256(file.body) : null,
+      bodySize: file.body?.byteLength ?? 0,
+      bodyText: bodyTextOrNull(file.mimeType, file.body),
+      ingestionConfig: file.ingestionConfig,
+    }))
+    .sort((a, b) => a.key.localeCompare(b.key));
```
Ostrzy: `sortBy`

services/confluence-connector/test/integration/scenario-context/unique-state.ts
```diff
@@ -0,0 +1,106 @@
...
+    .map((file) => ({
+      id: file.id,
+      key: file.key,
+      byteSize: file.byteSize,
+      mimeType: file.mimeType,
+      metadata: file.metadata,
+      bodyHash: file.body ? sha256(file.body) : null,
+      bodySize: file.body?.byteLength ?? 0,
+      bodyText: bodyTextOrNull(file.mimeType, file.body),
+      ingestionConfig: file.ingestionConfig,
```
Ostrzy, line 64: Shouldn't you have the scopeId here as well? If we don't shouldn't we have a scenario that tests that files land in correct spaces in knowledge base

services/confluence-connector/test/integration/scenario-context/unique-state.ts
```diff
@@ -0,0 +1,106 @@
...
+  const byId = new Map(scopes.map((scope) => [scope.id, scope]));
+  const result = new Map<string, string>();
+  for (const scope of scopes) {
+    const segments: string[] = [];
+    let current: { id: string; name: string; parentId: string | null } | undefined = scope;
+    const seen = new Set<string>();
+    while (current) {
+      if (seen.has(current.id)) {
+        break;
+      }
```
Ostrzy: It's a tree, we no cycles

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+export class FakeUniqueApi implements UniqueApiClient {
+  public readonly auth: UniqueApiClient['auth'];
+  public readonly scopes: UniqueApiClient['scopes'];
+  public readonly files: UniqueApiClient['files'];
+  public readonly users: UniqueApiClient['users'];
+  public readonly groups: UniqueApiClient['groups'];
+  public readonly ingestion: UniqueApiClient['ingestion'];
+  public readonly content: UniqueApiClient['content'];
+
+  private readonly scopesById = new Map<string, Scope>();
```
Ostrzy, line 57: Maybe we can just have an extended type here that also have the path so we don't do some funny logic when building state for comparison

services/confluence-connector/test/integration/scenario-context/unique-state.ts
```diff
@@ -0,0 +1,106 @@
...
+    return null;
+  }
+  if (!TEXT_MIME_TYPES.has(mimeType)) {
+    return null;
+  }
+  if (body.byteLength > SMALL_BODY_LIMIT_BYTES) {
+    return null;
+  }
+  return body.toString('utf-8');
+}
```
Ostrzy, line 80: I'm not sure what it does - we already have the buffer in memory, so it should be fine? Ask Your Friend Opus why it added it here

services/confluence-connector/test/integration/scenario/builders.ts
```diff
@@ -0,0 +1,161 @@
...
+  ScenarioSpace,
+  ScenarioUniqueFile,
+  ScenarioUniqueScope,
+} from './scenario.types';
+
+const DEFAULT_VERSION = '2026-05-01T10:00:00.000Z';
+const DEFAULT_SPACE_KEY = 'SP';
+const DEFAULT_SPACE_ID = 'space-1';
+const DEFAULT_TENANT_NAME = 'tenant1';
+const DEFAULT_INGEST_LABEL = 'ai-ingest';
```
Ostrzy: You said yourself to maybe export and use it when we rely on defaults

services/confluence-connector/test/integration/fakes/fake-blob-storage.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+function extractToken(path: string): string {
```
Ostrzy, line 64: I think that would be nice to explain in a comment what we do here and why because it's not obvious

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
```diff
@@ -0,0 +1,280 @@
...
+ */
+export class FakeConfluenceApi extends ConfluenceApiClient {
+  private readonly failures: FailureMap = {
+    getPageById: new Map(),
+    getAttachmentDownloadStream: new Map(),
+  };
+
+  public constructor(
+    private readonly tenant: ScenarioTenantConfig,
+    private readonly state: ScenarioConfluence,
```
Ostrzy: Please run it through `deepCopy` so we are sure the mutations are not by accident mutation default objects / arrays

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
```diff
@@ -0,0 +1,280 @@
...
+  public removeAttachment(pageId: string, attachmentId: string): void {
+    const page = this.requirePage(pageId);
+    const index = page.attachments?.findIndex((a) => a.id === attachmentId) ?? -1;
+    if (index === -1) {
+      throw new Error(`Cannot remove unknown attachment "${attachmentId}" from page "${pageId}"`);
+    }
+    page.attachments?.splice(index, 1);
+  }
+
+  // ─── Failure-injection hooks ─────────────────────────────────────────────────
```
Ostrzy: For the failure we can just mock it for the test

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
```diff
@@ -0,0 +1,280 @@
...
+    this.failures.getDescendantPages = undefined;
+    this.failures.getPageById.clear();
+    this.failures.getAttachmentDownloadStream.clear();
+  }
+
+  private requirePage(pageId: string): ScenarioPage {
+    const page = this.state.pages.find((p) => p.id === pageId);
+    if (!page) {
+      throw new Error(`Unknown page "${pageId}"`);
+    }
```
Ostrzy: `assert`

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  public clearFailures(): void {
+    this.failures.searchPagesByLabel = undefined;
+    this.failures.getDescendantPages = undefined;
+    this.failures.getPageById.clear();
+    this.failures.getAttachmentDownloadStream.clear();
+  }
+
+  private requirePage(pageId: string): ScenarioPage {
```
Ostrzy: `getPageOrFail`? 🤔

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+   *
+   * Whether a file is "moved" is decided server-side by Unique (same logical
+   * resource re-keyed to a new location), so the per-`partialKey` diff this fake
+   * reimplements cannot derive it on its own. This seam lets a test inject that
+   * server verdict, mirroring the `failOn*` injection hooks, so the connector's
+   * handling of moved files can be exercised end-to-end.
+   */
+  private readonly movedItemKeys = new Set<string>();
+
+  public constructor(private readonly initial: ScenarioUnique) {
```
Ostrzy: We don't have to `private readonly` this, we use it only in constructor

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+  public failOnPerformFileDiff(error: Error): void {
+    this.failures.performFileDiff = error;
+  }
+
+  /**
+   * Make the next `performFileDiff` report the given item keys as `movedFiles`
+   * instead of new/updated/deleted. See {@link movedItemKeys} for why this is
+   * injected rather than derived.
+   */
+  public simulateMovedFiles(itemKeys: string[]): void {
```
Ostrzy: This seems like quite a smell, we should set the state in a way that accomplishes that instead

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+  /** Test-only hook to seed a scope with a fixed id (e.g. the configured root scope). */
+  public seedScope(scope: Scope): void {
+    this.scopesById.set(scope.id, { ...scope });
+  }
+
+  public listFiles(): StoredFile[] {
+    return [...this.filesById.values()];
+  }
+
+  // ─── Failure-injection hooks ─────────────────────────────────────────────────
```
Ostrzy: The same here - we should mock errors instead of doing that

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+  private requireScope(scopeId: string): Scope {
+    const scope = this.scopesById.get(scopeId);
+    if (!scope) {
+      throw new Error(`Unknown scope: ${scopeId}`);
+    }
+    return scope;
+  }
+}
+
+function stringifyMetadata(
```
Ostrzy, line 391: Not sure what it actually brings over `JSON.stringify`

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -0,0 +1,473 @@
...
+  }
+
+  private finalizeIngestion(request: IngestionFinalizationRequest): { id: string } {
+    const failure = this.failures.finalizeIngestion.get(request.key);
+    if (failure) {
+      throw failure;
+    }
+    const file = [...this.filesById.values()].find((f) => f.key === request.key);
+    if (!file) {
+      throw new Error(`Cannot finalize unknown key "${request.key}"`);
```
Ostrzy: `assert`

services/confluence-connector/test/integration/scenario/scenario.types.ts
```diff
@@ -0,0 +1,105 @@
...
+  labels: string[];
+  parentId?: string;
+  versionWhen: string;
+  attachments?: ScenarioAttachment[];
+  /**
+   * Confluence content type. Defaults to `page`. Use `blogpost` to test blog
+   * post ingestion, or `database` / `whiteboard` / `embed` to test that the
+   * scanner skips these types (descendants are still discoverable via
+   * `parentId`, which is exercised by `content-types.integration-spec.ts`).
+   */
```
Ostrzy: THIS LOOKS LIKE SSHSHSHSHSHSHSHSLOP

services/confluence-connector/test/integration/scenario/scenario.types.ts
```diff
@@ -0,0 +1,105 @@
...
+  mimeType: string;
+  metadata?: Record<string, string>;
+  body?: Buffer;
+  /**
+   * ISO timestamp the file was last updated in Unique.
+   * Compared against the page's `versionWhen` by `performFileDiff` to decide
+   * whether the file is up-to-date (no change), older (updated), or missing (new).
+   * Defaults to epoch zero so any real page wins by default — set this when you
+   * want to test "no-change" or "older Unique data" scenarios.
+   */
```
Ostrzy: Let's remove these comments, we can add them to the scenarios instead to be clear how we handle it or maybe to methods that implement it like `performFileDiff`

services/confluence-connector/test/integration/scenarios/three-pages-one-space.scenario.ts
```diff
@@ -0,0 +1,19 @@
+import { aPage, aSpace } from '../scenario/builders';
+import { defineScenario } from '../scenario/scenario.builder';
+
+/**
+ * Confluence has three labeled pages in one space; Unique starts empty.
+ *
+ * Useful for failure-injection tests where one page fails and we want to
+ * verify the other two still ingest successfully.
+ */
+export const threePagesOneSpaceScenario = defineScenario({
```
Ostrzy: It's small enough and we use it in two scenarios, maybe not worth extracting 🤷🏼

services/confluence-connector/test/integration/__tests__/attachments.integration-spec.ts
```diff
@@ -0,0 +1,274 @@
...
+      },
+    });
+    ctx = buildScenarioContext(scenario);
+
+    const result = await ctx.runSync();
+
+    expect(result).toEqual({ status: 'success' });
+
+    const state = getUniqueState(ctx.unique);
+    const filesByKey = new Map(state.files.map((file) => [file.key, file]));
```
Ostrzy: We can just use `find`

services/confluence-connector/test/integration/__tests__/attachments.integration-spec.ts
```diff
@@ -0,0 +1,274 @@
...
+
+    expect(filesByKey.get('tenant1/space-1_SP/p1::att-png')).toMatchObject({
+      mimeType: 'image/png',
```
Ostrzy, line 72: Maybe a little bit more, like check id and name as well?

services/confluence-connector/test/integration/__tests__/attachments.integration-spec.ts
```diff
@@ -0,0 +1,274 @@
...
+      },
+    });
+    ctx = buildScenarioContext(scenario);
+
+    const result = await ctx.runSync();
+
+    expect(result).toEqual({ status: 'success' });
+
+    const state = getUniqueState(ctx.unique);
+    expect(state.files.map((file) => file.key).sort()).toEqual([
```
Ostrzy: Maybe we would like to have an expecter that takes a list of pages and attachments and checks if they landed in unique. Similarly one that checks they did not land in unique.

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  private toConfluencePage(
+    page: ScenarioPage,
+    options: { includeBody?: boolean } = {},
+  ): ConfluencePage {
+    const space = this.requireSpace(page.spaceKey);
+    const includeBody = options.includeBody ?? false;
+
+    const attachments: ConfluenceAttachment[] = (page.attachments ?? []).map((attachment) => ({
```
Ostrzy: Maybe it should limit to 25 like Confluence does

services/confluence-connector/test/integration/__tests__/subtree-sync.integration-spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    expect(result).toEqual({ status: 'success' });
+
+    const state = getUniqueState(ctx.unique);
+    expect(state.files.map((file) => file.key).sort()).toEqual([
+      'tenant1/space-1_SP/child',
+      'tenant1/space-1_SP/root',
+      'tenant1/space-1_SP/standalone',
+    ]);
+  });
```
Ostrzy, line 195: 'Maybe we need a case when you have `ai-ingest` page as child of not ingested page type, just to be sure that we have this covered

services/confluence-connector/test/integration/__tests__/update-content.integration-spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      bodyText: '<p>Updated content</p>',
+    });
+  });
+
+  // Full-replacement carve-out: an editor archived old pages (p1, p2) and
+  // republished new ones (p3, p4) with fresh page IDs. Every existing key
+  // would be deleted, but every key is also being replaced by a non-
+  // overlapping new one — a legitimate republication, not a bug. The guard
+  // logs a warning and lets the sync proceed.
+  it('allows a full replacement when every old key is replaced by a non-overlapping new key', async () => {
```
Ostrzy, line 91: We should also add a case when we block because some keys overlap

services/confluence-connector/test/integration/__tests__/root-scope-migration.integration-spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    // The new root has been claimed and now hosts the freshly ingested page.
+    const newRoot = state.scopes.find((scope) => scope.id === 'new-root');
+    expect(newRoot?.externalId).toBe('confc:cloud:cloud-1');
+    expect(newRoot?.path).toBe('/Confluence v2');
+
+    // The space scope was created under the new root.
+    expect(state.scopes.find((scope) => scope.path === '/Confluence v2/SP')).toBeDefined();
+
+    // Fresh content was ingested under the new root's space.
```
Ostrzy: It should've been moved. SHould we maybe check that no ingestion endpoint was called?

services/confluence-connector/test/integration/__tests__/root-scope-migration.integration-spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    expect(newKeys).toContain('tenant1/space-legacy_LEGACY/legacy-page');
+  });
+
+  // Conflict path: the operator pointed the tenant at a new root but forgot
+  // to clear the old root's externalId. The expected externalId is therefore
+  // already taken, so the new root's claim fails — and the entire sync fails
+  // safely without touching any data. The operator gets a clear signal that
+  // they must run the tenant-deletion flow against the old root before
+  // proceeding.
+  it('aborts the sync when the previous root scope still holds the expected externalId', async () => {
```
Ostrzy: No longer the case it seems, we clear when moving

services/confluence-connector/test/integration/__tests__/single-page-sync.integration-spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    expect(result).toEqual({ status: 'success' });
+
+    const state = getUniqueState(ctx.unique);
+    expect(state.files.map((file) => file.key)).toEqual(['tenant1/space-1_SP/labeled']);
+  });
+});
+
+function sha256(buf: Buffer): string {
+  return createHash('sha256').update(buf).digest('hex');
```
Ostrzy: We already have it somewhere else - make a tiny util

services/confluence-connector/test/integration/__tests__/delete-tenant.integration-spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    // "cleanup completed" marker.
+    expect(state.scopes).toEqual([
+      expect.objectContaining({ path: '/Confluence', externalId: null }),
+    ]);
+    expect(state.files).toEqual([]);
+  });
+
+  // The externalId on the root scope acts as the cleanup-completed marker:
+  // null after a successful run, set otherwise.
+  it('clears the root externalId only after the cleanup succeeds', async () => {
```
Ostrzy, line 92: Would be nice to have a case of run once, failed, we check the externalId is not removed and then we run again and succeed

services/confluence-connector/test/integration/__tests__/moved-content.integration-spec.ts
```diff
@@ -0,0 +1,67 @@
...
+ * The new / updated / deleted paths are covered by `single-page-sync`,
+ * `update-content`, and `delete-content` respectively.
+ */
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { aPage, aPageFile, aSpace, aSpaceScope } from '../scenario/builders';
+import { defineScenario } from '../scenario/scenario.builder';
+import { buildScenarioContext, type ScenarioContext } from '../scenario-context/scenario-context';
+import { getUniqueState } from '../scenario-context/unique-state';
+
+describe('moved content', () => {
```
Ostrzy: I don't get it, we don't ever move content I believe

Ostrzy (changes requested):

services/confluence-connector/test/integration/scenario/builders.ts
lorand93: c96475478487ed09abadd173e6b223168f0afab9

services/confluence-connector/test/integration/scenario/builders.ts
lorand93: 6816cd11979eac17d707fe92db9bc904728f5ded

services/confluence-connector/test/integration/scenario-context/scenario-context.ts
lorand93: as we've looked at this together:
in tenant-sync.scheduler.ts when tenant.status === Deleted the scheduler calls deleteService.deleteTenantContent() and returns early, it never calls synchronize()

I've updated the description of the test so explicitly mention this.

services/confluence-connector/test/integration/fakes/fake-blob-storage.ts
lorand93: dead weight indeed, removing.

lorand93: c5b091d755c426bd958a0f5cb776c7fe0be6236d

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
lorand93: removed, removed the optionality in the typing also. All scenarios always set updatedAt

lorand93: ae3c4b27cd9071dda832d3746735c2191df36a37

services/confluence-connector/test/integration/scenario-context/scenario-context.ts
lorand93: defaulted the root scope in defineScenario, removed the seedRootScopeIfMissing case from scenario-context.ts.

  2d1be01eb20314e65304a56efdc63544d547e754

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
lorand93: true :D debd8c916aa4acb4cd1bb9db4350b2de98945a85

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
lorand93: 0f0fffa2167dd62371d22cf47410ef521375fe14

services/confluence-connector/test/integration/scenario-context/unique-state.ts
lorand93: 608bd18f63687951f6c02177bb83288a38e02c96

services/confluence-connector/test/integration/scenario-context/unique-state.ts
lorand93: 608bd18

services/confluence-connector/test/integration/scenario-context/unique-state.ts
lorand93: yup 6e18fce525c06913eaf37a8cdc2480dac0a4168b

services/confluence-connector/test/integration/scenario-context/unique-state.ts
lorand93: 4780def7ee62540b91c38b15388c13b95dd4e889

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
lorand93: added scopeWithPath that extends Scope and removed the helper function to compute the scope path. 

1fa4df2e124d6079a21727082ce294744c3abd4b

services/confluence-connector/test/integration/scenario-context/unique-state.ts
lorand93: garbage, removed

services/confluence-connector/test/integration/scenario/builders.ts
lorand93: 4981ef59082b0f155b5b4c376d2d49d917ae4d96

6927184c68c987a50628ba0e02c48b6a2506c719

services/confluence-connector/test/integration/fakes/fake-blob-storage.ts
lorand93: 634a75c6a1748e377077023a4df5046fba851cbb

7aaa5318e4977bc4cd0a542227703316110d89e4
f6f16fe7bc5ba5b5fdfb8a6146296f5d54512f4f

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
lorand93: interesting eb9c88a68dac3f8b9cb8affc000963ed0382b93a

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
lorand93: 70da0d47fb1be3e98c601313a46fc7fc0623001c

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
lorand93: 3bf24f5f23005de63c16184729dfb9cd1a4e6c00

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
lorand93: > a8b05333db8124b40dd18c09d5f0656350ef2f99

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
lorand93: replaced with return mapValues(metadata, (value) =>
      388 +    typeof value === 'string' ? value : JSON.stringify(value),
      389 +  );

lorand93: 19baf62ce9e26201e262b82ed35092868a9ec677

services/confluence-connector/test/integration/scenario/scenario.types.ts
lorand93: 45c31693

services/confluence-connector/test/integration/scenario/scenario.types.ts
lorand93: 952fb9fe92c85e6035344f690681379501aa6f60

services/confluence-connector/test/integration/__tests__/attachments.integration-spec.ts
lorand93: 59dfd3b748d474f95ea4f45bce20fbf8ff08b3f9

services/confluence-connector/test/integration/scenarios/three-pages-one-space.scenario.ts
lorand93: 513f29910518bf4f26c648305826cf4a5d75754b

services/confluence-connector/test/integration/__tests__/attachments.integration-spec.ts
lorand93: 3b670a17dda23c11bec8f02359cb94b94f6d7127

services/confluence-connector/test/integration/__tests__/subtree-sync.integration-spec.ts
lorand93: 7c8d8dbf2394bb496a63dbc8a05cc107bd16ffe6

services/confluence-connector/test/integration/fakes/fake-confluence-api.ts
lorand93: 7116e3f80f5301e822cc70a0f0bc1cf090e155e9

services/confluence-connector/test/integration/__tests__/attachments.integration-spec.ts
lorand93: f887e4d40e09bb888e42ee38d652e555c08b4f2f
dfbfda519069f94b8b560cb30fc2e222914a9e7c
706ae43f43d864bd6b17d75d96e59cd06bb0db24

services/confluence-connector/test/integration/__tests__/update-content.integration-spec.ts
lorand93: 75cbe0e4923b8f7c6a536b72cd54849326b92c48

services/confluence-connector/test/integration/__tests__/root-scope-migration.integration-spec.ts
lorand93: 53042850051364a87d3eabe66f96d3195825c5d1

services/confluence-connector/test/integration/__tests__/root-scope-migration.integration-spec.ts
lorand93: merged main to this branch and adjusted tests

f4f7b41b581ed75806711394469453f82a1c9b06

services/confluence-connector/test/integration/__tests__/single-page-sync.integration-spec.ts
lorand93: no need for this anymore, deleted 

f86a0ef1a241fe95b5cef123af4868f6efe76688

services/confluence-connector/test/integration/__tests__/delete-tenant.integration-spec.ts
lorand93: 3b7a540cbaa243f3d534ff64cb13350a9dc1e233

services/confluence-connector/test/integration/__tests__/moved-content.integration-spec.ts
lorand93: true! Replaced with cross-space move

lorand93: 9f1ca08b269667389534051edf07b87bb9be5c59

lorand93:
bugbot review

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
```diff
@@ -394,74 +334,69 @@ export class FakeUniqueApi implements UniqueApiClient {
...
-    const existing = [...this.filesById.values()].filter((f) => f.key.startsWith(`${partialKey}/`));
+    const prefix = `${partialKey}/`;
+
+    // The diff is scoped to a single partial key, mirroring Unique's per-space
+    // file-diff: only content stored under this partial key is compared against
+    // the submitted items. Content under other partial keys is invisible here,
+    // so a page that exists under a different space is simply "new" to this
+    // space (a cross-space move surfaces as new-here plus deleted-there, not as
+    // a single move).
+    const here = new Map<string, StoredFile>();
```
Ostrzy: It's a slightly weird variable name 😅

Ostrzy (approved):

services/confluence-connector/test/integration/fakes/fake-unique-api.ts
lorand93: indeed dc8cff74288c579536f940e629cd0b1c106182f6

PR #528, opened 2026-05-11
feat(sharepoint-connector): support `in_parent:` scope auto-creation

Ostrzy:
Accept `in_parent:scope_<id>` as a SharePoint List `scopeId` value; connector finds-or-creates a child scope under the given parent and uses it as the site root, mirroring v1's auto-create UX without requiring Champions to pre-create a scope per site.

- **Resolution**: try `getScopeByExternalId('spc:{siteId}/site')` and move on parent-mismatch; otherwise scan one `listChildrenScopes(parent)` call for both legacy-externalId reuse and the name-match abort cases (unclaimed / foreign / ambiguous); finally create + claim, rolling back the create if claiming fails. Existing `ScopeExternalIdMigrationService` (running inside `initializeRootScope`) normalizes any legacy externalId afterward.
- **Unify site deletion**: `auto` rows additionally delete the auto-created root after `resetRootScope`; `fixed` rows behave as today. The parent's externalId is never written.
- **Dedup**: keep dedup-by-scopeId for `fixed` rows (skipping `auto`, since multiple sites legitimately share a parent) and add dedup-by-siteId so the same SharePoint site cannot be configured to two different roots.
- No helm / Terraform / RBAC changes; docs updated for the new prefix.

Ticket: UN-20365


services/sharepoint-connector/src/sharepoint-synchronization/root-scope/create-root-scope.command.ts
```diff
@@ -0,0 +1,150 @@
...
+        siteName,
+        detail: `failed to set externalId on newly-created scope ${created.id}`,
+        cause: claimError,
+      });
+    }
+  }
+
+  private async rollbackCreatedScope(scopeId: string, logPrefix: string): Promise<void> {
+    try {
+      await this.uniqueScopesService.deleteScope(scopeId, { recursive: true });
```
nicolae536, line 142: One thing -> after 30 seconds the connection is cut and delete continues in the background based on how much content you have this can be super slow -> I think you are better off paginating content wich has `folderIdPath` containing your scope id and delete in batches, this if this might be an issue or not if I get it correctly scopes are mostly empty in this cases

services/sharepoint-connector/src/sharepoint-synchronization/root-scope/initialize-root-scope.command.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    return { rootScopeId, serviceUserId: userId, rootPath, isInitialSync };
+  }
+
+  private async resolveRootScopeId(
+    siteConfig: SiteConfig,
+    siteName: Smeared,
+  ): Promise<{ rootScopeId: string; precomputedRootPath: Smeared | undefined }> {
+    if (isFixedScope(siteConfig.scopeId)) {
+      return { rootScopeId: siteConfig.scopeId.scopeId, precomputedRootPath: undefined };
```
nicolae536, line 124: siteConfig.scopeId.scopeId this naming is totally funny

services/sharepoint-connector/src/sharepoint-synchronization/root-scope/resolve-scope-path.command.ts
```diff
@@ -0,0 +1,35 @@
...
+export class ResolveScopePathCommand {
+  public constructor(private readonly uniqueScopesService: UniqueScopesService) {}
+
+  // Walks up from `scope` to the root, granting READ permission to `userId` on each ancestor
+  // first so that `getScopeById` can read it. Returns the absolute path including `scope.name`.
+  public async execute(scope: Scope, userId: string): Promise<Smeared> {
+    const pathSegments = [scope.name];
+    let currentScope: Scope = scope;
+
+    while (currentScope.parentId) {
```
nicolae536, line 17: I think you can do this with 1 call -> if you look in `metadata.folderIdPath` you probably have all ids there and then you create all of them at once.

services/sharepoint-connector/src/sharepoint-synchronization/scope-management.service.ts
```diff
@@ -3,144 +3,33 @@ import { randomUUID } from 'node:crypto';
...
-      }
-    }
-
-    const pathSegments = [rootScope.name];
-    let currentScope: Scope = rootScope;
-
-    while (currentScope.parentId) {
-      // Grant READ permission first before accessing the parent scope. Otherwise we will not get it
-      // via `getScopeById` call.
-      await this.uniqueScopesService.createScopeAccesses(currentScope.parentId, [
```
nicolae536, line 124: ah now I see this was just splitted

nicolae536 (approved):

services/sharepoint-connector/src/sharepoint-synchronization/root-scope/resolve-scope-path.command.ts
nicolae536: My bad this on Content not on scope

services/sharepoint-connector/src/sharepoint-synchronization/root-scope/create-root-scope.command.ts
Ostrzy: As we discussed, that should be always empty when this deletion happens 🙂

services/sharepoint-connector/src/sharepoint-synchronization/scope-management.service.ts
Ostrzy: Yup, scope manegement service got way to big to work with it 😅

services/sharepoint-connector/src/sharepoint-synchronization/root-scope/initialize-root-scope.command.ts
Ostrzy: I know, but I don't know what would be the sensible naming here, because sometimes it's that, other times it's the parent ID 😅 I guess small downside from re-using the config to handle the `in_parent:` case, but I think a sensible price 🙂

PR #546, opened 2026-05-13
feat(kyckr-mcp): download document as json format

lorand93:
## Summary

Reshapes how `create_document_order` and `get_order` deliver the ordered
document body so the agent presents contents to the user instead of
pasting raw `/orders/:id/download` URLs.

- JSON-first, PDF-fallback on the same order. On any `Success` status, `fetchOrder` tries `/orders/:id/download?format=json` and inlines the parsed body under `data.documentJson`. If the registry has no JSON projection (`Data: null`, empty body, 404), the same order is re-requested as `?format=pdf` (no second order is placed) and the PDF (≤ 8 MB) is attached to the tool result as an MCP `EmbeddedResource` block. Larger PDFs return a size note instead.

## Caveats

- Client passthrough of `EmbeddedResource` is required for the LLM to see the PDF; verified in Cursor. Other MCP clients vary.
- 8 MB raw PDF cap (~11 MB base64). Tunable via `MAX_PDF_BYTES` in `_shared/fetch-order.ts`.
- Mixed-content path bypasses `mcp-server-module`'s auto-validation of `outputSchema`; queries validate manually before returning.


lorand93:
bugbot review

Ostrzy (approved):

PR #563, opened 2026-05-19
feat(confluence-connector,deps,main): inline page image attachments as base64 data URIs

lorand93:
Closes UN-14316 (UN-20384).

When the connector ingests a Confluence page that references image attachments, the images are now downloaded and embedded as base64 `data:` URIs directly inside the page's HTML. A page with images becomes a single self-contained ingestion artifact instead of one page + N separate image artifacts. When inlining is enabled, image attachments are excluded from the standalone attachment path entirely: orphan images (attached but not referenced by an in-body macro) are appended to the end of the page body, and a macro-referenced image that cannot be inlined keeps its original `<ac:image>` macro and is not ingested elsewhere. Non-image attachments still flow through the standalone attachment path unchanged.

## Configuration

Inlining is turned on by the presence of the platform's image-extraction config on the connector: set `ingestion.pageIngestionConfig.htmlConfig.imageContentExtraction.languageModel` to the visual LLM configured in Knowledge Base. Its presence is the toggle (there is no separate `inlineImages` flag). When it is absent, inlining is off and image attachments are ingested as standalone attachments instead, which is the right setup for platforms older than `2026.24.0` that cannot extract text from inline base64 images.

```
Example from confluence page:
<ac:image ac:align="center" ac:layout="center" ac:original-height="275"
            ac:original-width="816" ac:custom-width="true" ac:local-id="5d5f924a09fa"
            ac:alt="asdasdasdas dasdasd as" ac:width="760">
    <ri:attachment ri:filename="image-20260428-105243.png" ri:version-at-save="1" />
    <ac:caption ac:local-id="81d1a65871bc">
      <p>asdasdasdasd as das da sdas da sda sda </p>
    </ac:caption>
  </ac:image>

  …and:

  <ac:image ac:align="center" ac:layout="center" ac:original-height="505"
            ac:original-width="583" ac:custom-width="true" ac:local-id="3d1dc5568905"
            ac:alt="image-20260428-105253.png" ac:width="583">
    <ri:attachment ri:filename="image-20260428-105253.png" ri:version-at-save="1" />
  </ac:image>
  ```
## New flow

```mermaid
flowchart TB
  Discover["Discover labeled pages<br/>+ attachments"]
  Diff["Compute file diff<br/>(image attachments excluded when inlining is on)"]
  EnsurePagesLoop{{"For each new/updated page"}}
  Fetch["Fetch page body<br/>(Confluence storage XML)"]
  Inline["Inline image attachments<br/>(parse &lt;ac:image&gt;, download, base64-encode,<br/>splice &lt;img src='data:...'&gt;; append orphan images)"]
  IngestPage["Register + upload + finalize<br/>(single page artifact with inlined images)"]
  EnsureAttLoop{{"For each remaining attachment<br/>(non-images only)"}}
  IngestAtt["Register + stream upload + finalize"]
  Delete["Delete removed items"]

  Discover --> Diff
  Diff --> EnsurePagesLoop
  EnsurePagesLoop --> Fetch
  Fetch --> Inline
  Inline --> IngestPage
  IngestPage --> EnsurePagesLoop
  EnsurePagesLoop --> EnsureAttLoop
  EnsureAttLoop --> IngestAtt
  IngestAtt --> EnsureAttLoop
  EnsureAttLoop --> Delete
```

Asked opus to sum up the feature:
```
Discovery. During the normal page scan, the connector already fetches each labeled page's attachments via
  expand=children.attachment. For every image attachment (mediaType starts with image/), we bucket it by page id into a pageId →
  image attachments map.

  Pages pass. For each new/updated page:

  1. Fetch the page body — Confluence storage XML, contains <ac:image> macros.
  2. Parse the body with parseImageBlocks (htmlparser2 + DOM walk). Each <ac:image> becomes a record with its byte range in the
  source and its resource ref: current-page attachment, cross-page attachment (<ri:page> child pointing to another page), or
  external URL (<ri:url>).
  3. Resolve each block to attachment metadata:
    - Current-page: match by filename in the discovered attachments for this page.
    - Cross-page: call fetchPageAttachmentsByTitle(spaceKey, title) and match. Per-sync cache so two refs to the same target page
  only hit the API once.
    - External URL or unresolved: skip, leave the macro as-is.
  4. Filter each resolved attachment: must be image/*, must be in allowedMimeTypes, must be ≤ maxFileSizeMb. Any failure → leave
  the macro untouched.
  5. Download the binary through the existing rate-limited Confluence client.
  6. Build a replacement <img src="data:image/<type>;base64,<...>" alt="…" width="…" height="…" /> — alt/title/width/height carried
   over from the original <ac:image> attributes.
  7. Splice all replacements into the original body by byte range. Content outside the swapped regions is preserved byte-for-byte.
  8. Append any of the page's image attachments that no in-body macro referenced (orphans) to the end of the body.
  9. Upload the enriched body as the single page artifact (registerContent + buffered PUT).

  Attachments pass. When inlining is enabled, image attachments are excluded from the standalone-attachment list up front
  (filtered by MIME type), so only non-image attachments run through the existing ingestAttachment path. So:

  - Images referenced by an in-body macro → inlined into the page HTML.
  - Orphan images (attached but referenced by no page body) → appended to the end of the page body.
  - Images that fail to inline (download error, MIME outside the allowlist, oversize, target page not found) → macro left
  untouched, not ingested elsewhere. The download uses the same rate-limited Confluence client as the standalone path, so a
  failure means the attachment is unreachable; a transient failure is retried on the next sync once the page re-ingests.
  - Non-image attachments (PDF, DOCX, etc.) → standalone path, unchanged.

  Safety nets baked in:

  - Cross-page lookup cache is reset at the start of each sync, so attachment changes are picked up next cycle.
  - Cross-page lookup errors are evicted from the cache so a transient blip doesn't poison the whole sync; legitimate "page not
  found" stays cached.
  - Malformed <ac:image> macros (unclosed, etc.) are skipped via a literal endsWith('</ac:image>') / '/>' check — they can't splice
   into unrelated content.
  - If the inliner throws unexpectedly mid-page, the orchestrator catches it and uploads the original body so the page is never
  lost.
  ```
## Image-reference handling

| Storage-format shape | Action |
|---|---|
| `<ac:image><ri:attachment ri:filename="..."/></ac:image>` | Inlined from the current page's attachments |
| `<ac:image><ri:attachment ri:filename="..."><ri:page ri:space-key="..." ri:content-title="..."/></ri:attachment></ac:image>` | Resolved via Confluence API (cached per sync) and inlined |
| `<ac:image><ri:url ri:value="https://..."/></ac:image>` | Left untouched in the HTML, never fetched |

## Test plan

- [ ] Local sync against `local-confluence-v8.5-docker-compose.yaml` with a page containing an inline image attachment and a cross-space reference. Verify the registered page body contains `` for both images, no `<ac:image>` macros remain, no standalone artifacts for the inlined images.
- [ ] Same as above but with an orphan image attachment on the page (no body reference). Verify it is appended to the page body as a base64 `` and is not ingested as a standalone artifact.
- [ ] A page that references a missing/oversize/external image. Verify the corresponding `<ac:image>` macro is preserved, the image is not ingested standalone, and the connector logs the reason.

### CODE REVIEW

**Comment by lorand93** [comment]
bugbot review

File: services/confluence-connector/src/synchronization/confluence-synchronization.service.ts [hunk 1]
```diff
@@ -251,3 +306,8 @@ export class ConfluenceSynchronizationService {
     return map;
   }
 }
+
+function isImageMediaType(mediaType: string): boolean {
+  const normalized = mediaType.split(';')[0]?.trim().toLowerCase() ?? '';
+  return normalized.startsWith('image/');
+}
```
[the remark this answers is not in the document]
lorand93: 68a9a3f8f0bc474c3f32eae2e972868468f60acf

lorand93:
bugbot review

services/confluence-connector/src/synchronization/confluence-tags-parser.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+  const page = firstChildElementByName(attachment, 'ri:page');
+  if (page) {
+    const spaceKey = page.attribs['ri:space-key'];
+    const contentTitle = page.attribs['ri:content-title'];
+    if (spaceKey && contentTitle) {
+      return { kind: 'cross-page-attachment', filename, spaceKey, contentTitle };
+    }
+  }
+  return { kind: 'current-attachment', filename };
```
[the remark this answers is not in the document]
lorand93: true [527d3ee](https://github.com/Unique-AG/connectors/pull/563/commits/527d3ee372e9c4bec54717cdff3872b2dcacb1bb)

services/confluence-connector/package.json
```diff
@@ -42,6 +42,8 @@
     "@unique-ag/utils": "workspace:*",
     "bottleneck": "2.19.5",
     "cron": "4.3.5",
+    "domhandler": "5.0.3",
+    "htmlparser2": "10.1.0",
```
Ostrzy: I will never stop being paranoid about agents adding libs so I ask 😄 Did we check these versions are the latest and within 72h release policy? 🙂

services/confluence-connector/src/synchronization/media-type.ts
```diff
@@ -0,0 +1,9 @@
+export function normalizeMediaType(mediaType: string): string {
```
Ostrzy: Would be nice to have a comment what is the expected input and what we get out of it with couple examples

services/confluence-connector/src/synchronization/confluence-tags-parser.ts
```diff
@@ -0,0 +1,108 @@
...
+      startIndex: node.startIndex,
+      endIndex,
+      imgAttrs: node.attribs,
+      resourceRef: resolveResourceRefFromNodes(node),
+    });
+  }
+  return blocks;
+}
+
+function collectImageNodes(parent: ParentNode, out: Element[]): void {
```
Ostrzy: 1. Doesn't we have some kind of `querySelectAll` for this parser?
2. We could return the array of elements instead of mutating single one, no?

services/confluence-connector/src/synchronization/page-image-inliner.ts
```diff
@@ -0,0 +1,380 @@
...
+    string,
+    Promise<PageAttachmentLookupResult | null>
+  >();
+
+  public constructor(
+    private readonly config: TenantConfig,
+    private readonly confluenceApiClient: ConfluenceApiClient,
+  ) {}
+
+  public resetOtherPageAttachmentCache(): void {
```
Ostrzy: Let's check why we need it, I believe this service is initialized once per tenant and execution, so the exact granularity we need.

services/confluence-connector/src/synchronization/page-image-inliner.ts
```diff
@@ -0,0 +1,380 @@
...
+  private buildAttachmentTitleMap(
+    attachments: DiscoveredAttachment[],
+  ): ReadonlyMap<string, DiscoveredAttachment> {
+    const attachmentsByTitle = new Map<string, DiscoveredAttachment>();
+    for (const attachment of attachments) {
+      if (!attachmentsByTitle.has(attachment.title)) {
+        attachmentsByTitle.set(attachment.title, attachment);
+      }
+    }
+    return attachmentsByTitle;
```
Ostrzy: A NIT: We could user indexBy from remeda

services/confluence-connector/src/synchronization/page-image-inliner.ts
```diff
@@ -0,0 +1,380 @@
...
+  return `${pageId}::${attachmentId}`;
+}
+
+export class PageImageInliner {
+  private readonly logger = new Logger(PageImageInliner.name);
+  // Caches "list attachments of page <title> in space <key>" lookups, used when an image
+  // macro references an attachment on a different page. Per-sync only; the orchestrator
+  // calls resetOtherPageAttachmentCache() at the start of each sync so attachment changes
+  // on referenced pages are picked up on the next run.
+  private readonly otherPageAttachmentCache = new Map<
```
Ostrzy: Regarding discussion we had - maybe we want just a reporting whether this happens a lot for our production, so we do not accidentally trip ourselves on memory

services/confluence-connector/src/synchronization/page-image-inliner.ts
```diff
@@ -0,0 +1,380 @@
...
+          err,
+          msg: 'Attachment lookup on referenced other page failed',
+        });
+        return null;
+      });
+    this.otherPageAttachmentCache.set(cacheKey, promise);
+    return promise;
+  }
+
+  private isAllowedMediaType(mediaType: string): boolean {
```
Ostrzy: I think it should be `MimeType` instead of `MediaType` everywhere 🙂

services/confluence-connector/src/synchronization/confluence-synchronization.service.ts
```diff
@@ -237,6 +276,24 @@ export class ConfluenceSynchronizationService {
...
+        continue;
+      }
+      const existing = map.get(attachment.pageId);
+      if (existing) {
+        existing.push(attachment);
+      } else {
+        map.set(attachment.pageId, [attachment]);
+      }
+    }
+    return map;
```
Ostrzy: NITPICK, feel free to ignore - this could be filter + groupBy

services/confluence-connector/src/synchronization/confluence-synchronization.service.ts
```diff
@@ -90,12 +97,25 @@ export class ConfluenceSynchronizationService {
...
+        const inlinedAttachmentIds = new Set<string>();
 
         this.metrics.setSyncPhase(SyncPhase.IngestingPages);
-        await this.fetchAndIngestPages(pagesToFetch, spaceScopes, concurrency);
+        await this.fetchAndIngestPages(
+          pagesToFetch,
+          spaceScopes,
+          concurrency,
+          imageAttachmentsByPageId,
+          inlinedAttachmentIds,
```
Ostrzy: We should probably return it instead of passing empty and expect it to be filled

services/confluence-connector/src/confluence-api/index.ts
```diff
@@ -1,3 +1,4 @@
+export type { PageAttachmentLookupResult } from './confluence-api-client';
```
Ostrzy: sus

services/confluence-connector/src/confluence-api/data-center-api-client.ts
```diff
@@ -85,6 +88,28 @@ export class DataCenterConfluenceApiClient extends ConfluenceApiClient {
     return result.success ? result.data : null;
   }
 
+  public async fetchPageAttachmentsByTitle(
+    spaceKey: string,
+    title: string,
```
Ostrzy: `pageTitle` and you can add a tiny comment that indeed it is unique across the space

services/confluence-connector/src/synchronization/page-image-inliner.ts
```diff
@@ -0,0 +1,380 @@
...
+      return { page, inlinedAttachmentIds: new Set() };
+    }
+
+    const blocks = parseImageBlocks(page.body);
+    if (blocks.length === 0) {
+      return { page, inlinedAttachmentIds: new Set() };
+    }
+
+    const pageImageAttachmentsByTitle = this.buildAttachmentTitleMap(pageImageAttachments);
+    const replacements = await Promise.all(
```
Ostrzy: Would be nice to batch these requests in here, by 20 or 50 or so

services/confluence-connector/docs/operator/configuration.md
```diff
@@ -302,11 +302,13 @@ These are matched against the `mediaType` reported by the Confluence API. Operat
 
 #### Image Attachments
```
Ostrzy, line 319: We need somewhere the compatibility section that this new version with inline images works only with Unique platform ver 2026.XX

services/confluence-connector/src/synchronization/confluence-tags-parser.ts
lorand93: 625462c7e8e47dff9268ee181ee0978a8332ca00

services/confluence-connector/src/synchronization/media-type.ts
lorand93: 6d352f903e935a36cdbe945d6486d4d5ab43fb9b

services/confluence-connector/package.json
lorand93: true thing. I updated them. 

image
image


```
 "domhandler": "6.0.1",
      46 +    "htmlparser2": "12.0.0",
```

lorand93: eb1957af85069fc83c7d32498b7d263bd5cf27b6

services/confluence-connector/src/synchronization/confluence-tags-parser.ts
lorand93: a06ad4e3dd0580ca64051c218b12cde9415e8ca9

services/confluence-connector/src/synchronization/page-image-inliner.ts
lorand93: 74d9e00d55e64fc8746c63f7da6f4ba4ccb476b1

services/confluence-connector/src/synchronization/page-image-inliner.ts
lorand93: changed. looks like so this bullshit naming comes from the confluence api / schema

services/confluence-connector/src/synchronization/page-image-inliner.ts
lorand93: It was caching only the metadata at this point, but I dropped the cache for now as it really doesn't make sense to have it for this edge case 4a4ad5a91c69fbdc7d76852024eecf336b958b35

services/confluence-connector/src/synchronization/confluence-synchronization.service.ts
lorand93: 1a47a7d80ca91c894c0aa0dedc4ab3e3ad8f805d

services/confluence-connector/src/synchronization/confluence-synchronization.service.ts
lorand93: this one was really ugly that I left in🤮🤮🤮 6823b9ac1151249441626d1629beca3a1f9e1e01

services/confluence-connector/src/confluence-api/index.ts
lorand93: consumed only by a test... (page-image-inliner.spec.ts:296) 

removed. dab44354feb6e7dc3f7ea058b800281ad6bbaaa5

services/confluence-connector/src/confluence-api/data-center-api-client.ts
lorand93: dab44354feb6e7dc3f7ea058b800281ad6bbaaa5 

b9045e97297cd1e31b9723fcfd27458003eecea6

Renamed for clarity

services/confluence-connector/src/synchronization/page-image-inliner.ts
lorand93: 7e6c71f420127d9bcacb10d74aa71ce6b4dbfa09

services/confluence-connector/docs/operator/configuration.md
lorand93: 4d5b1bd929f59a6c29c7afd270ab4949e8b0d31a
 
 True!

services/confluence-connector/src/synchronization/page-image-inliner.ts
lorand93: removed the cache entirely. cross-page references are rare enough that re-fetching is fine

lorand93:
bugbot review

services/confluence-connector/docs/technical/flows.md
```diff
@@ -110,12 +110,23 @@ sequenceDiagram
         loop For each new/updated page (concurrency-limited)
             Connector->>Confluence: GET page by ID<br/>(body.storage)
             Confluence->>Connector: Page HTML content
+            Note over Connector: Parse storage XML, locate &lt;ac:image&gt; macros
+            opt Image attachments referenced
+                opt Some references point to another page
+                    Connector->>Confluence: GET /rest/api/content?spaceKey=&amp;title=<br/>(cached per sync)
```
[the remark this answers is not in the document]
lorand93: fixed

services/confluence-connector/src/synchronization/page-image-inliner.ts
```diff
@@ -0,0 +1,326 @@
...
+      resolved.downloadPath,
+    );
+    const html = this.buildImgTag(macro.imgAttrs, resolved.mediaType, buffer, resolved.filename);
+    return {
+      start: macro.startIndex,
+      end: macro.endIndex,
+      attachmentId: resolved.attachmentId,
+      pageId: resolved.pageId,
+      html,
+    };
```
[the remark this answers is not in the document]
lorand93: fixed

services/confluence-connector/docs/operator/configuration.md
```diff
@@ -304,6 +304,8 @@ These are matched against the `mediaType` reported by the Confluence API. Operat
 
 Images embedded in Confluence pages (drag/drop, paste, or "Insert image") are stored as regular page attachments by Confluence. During page ingestion the connector inlines each referenced image directly into the page HTML as a base64 `data:` URI, producing a single self-contained page artifact rather than a separate image artifact. This applies to images attached to the page being ingested as well as images attached to other pages in the same Confluence instance (references to an attachment on another page). PNG and JPEG are the supported formats; both are in the default `allowedMimeTypes`.
 
+> **Platform compatibility:** Inline images require Unique platform `2026.24.0` or later. The platform extracts searchable text from inline base64 images starting with that version; on earlier versions the page is still ingested but its embedded images contribute no searchable content. Extraction is also gated behind two company-scoped feature flags that must be enabled (`FEATURE_FLAG_ENABLE_HTML_INLINE_IMAGE_EXTRACTION_UN_20936` and `FEATURE_FLAG_ENABLE_MULTI_FILE_IMAGE_CONTENT_EXTRACTION_UN_20936`). See [Unique Platform Compatibility](./deployment.md#unique-platform-compatibility).
```
Ostrzy: So if someone needs this or later version they need to upgrade the platform or the images will effectively disappear? 🤔 Hopefully that won't happen but if it does we may have to add a config later 🙂

Ostrzy (approved):

services/confluence-connector/docs/operator/configuration.md
lorand93: Yes, exactly. On versions before 2026.24.0 the old HTML ingestor drops base64 data: images, and since the connector inlines them it no longer ingests them as standalone attachments either, so the image content isn't searchable.

we can add an attachments.inlineImages: enabled|disabled toggle -> let me do that in this PR

lorand93:
bugbot review

services/confluence-connector/src/synchronization/confluence-synchronization.service.ts
```diff
@@ -62,15 +67,22 @@ export class ConfluenceSynchronizationService {
...
+      // we either inline images in the page, either ingest them separately with no inlining.
+      // when inlining we must exclude them from file-diff else we will ingest them twice: inlined and as attachment.
+      const trackedAttachments = tenant.config.ingestion.attachments.inlineImagesEnabled
+        ? discoveredAttachments.filter((a) => !isImageMimeType(a.mediaType))
+        : discoveredAttachments;
+
       const diffResult = await this.fileDiffService.computeDiff(
         discoveredPages,
-        discoveredAttachments,
+        trackedAttachments,
```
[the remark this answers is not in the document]
lorand93: attachment changes do bump page last-modified, so
  the diff re-ingests the page. No code change needed.

services/confluence-connector/src/synchronization/confluence-synchronization.service.ts
```diff
@@ -89,13 +101,25 @@ export class ConfluenceSynchronizationService {
...
+          imageAttachmentsByPageId,
+        );
 
         this.metrics.setSyncPhase(SyncPhase.IngestingAttachments);
-        await this.ingestAttachments(attachmentsToIngest, spaceScopes, concurrency);
+        await this.ingestAttachments(
+          attachmentsToIngest,
+          spaceScopes,
+          tenant.config.processing.concurrency,
+        );
```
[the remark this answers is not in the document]
lorand93: Inlining downloads through the same rate-limited Confluence client the standalone path uses, so a failure here means the attachment is unreachable. The standalone path would fail on the same call.

services/confluence-connector/docs/operator/configuration.md
```diff
@@ -272,6 +287,7 @@ ingestion:
 | `storeInternally` | No      | `enabled`        | Whether to store content internally in Unique (`enabled` or `disabled`)                                                                                             |
 | `useV1KeyFormat` | No       | `disabled`       | Use v1-compatible ingestion key format (`spaceId_spaceKey/pageId`) without tenant prefix (`enabled` or `disabled`). Only relevant when migrating from Confluence Connector v1 |
 | `attachments`    | No       | (see sub-fields) | Configuration for file attachment ingestion                                                                                                                          |
+| `pageIngestionConfig` | No | -- | Ingestion configuration applied to each ingested page. Setting `htmlConfig.imageContentExtraction.languageModel` to the visual LLM configured in Knowledge Base turns on inlining of page images as base64 `data:` URIs (see [Image Attachments](#image-attachments)). Without it, images fall back to standalone attachment ingestion |
```
Ostrzy: The link is wrong, it's in github style instead of confluence style 🙂

services/confluence-connector/docs/operator/configuration.md
```diff
@@ -302,13 +318,17 @@ These are matched against the `mediaType` reported by the Confluence API. Operat
...
 
-Image attachments fall back to standalone ingestion when they cannot be inlined: orphan images (attached but not referenced by any page body), images that exceed `attachments.maxFileSizeMb`, images whose download fails, and references to an attachment on another page whose referenced page cannot be resolved.
+> **Platform compatibility:** Inline images require Unique platform `2026.24.0` or later. The platform extracts searchable text from inline base64 images starting with that version; on earlier versions the page is still ingested but its embedded images are dropped during HTML-to-Markdown conversion and contribute no searchable content. Extraction is also gated behind two company-scoped feature flags that must be enabled (`FEATURE_FLAG_ENABLE_HTML_INLINE_IMAGE_EXTRACTION_UN_20936` and `FEATURE_FLAG_ENABLE_MULTI_FILE_IMAGE_CONTENT_EXTRACTION_UN_20936`). See [Unique Platform Compatibility](./deployment.md#unique-platform-compatibility).
```
Ostrzy: Wrong link here as well 🙂

services/confluence-connector/docs/operator/configuration.md
```diff
@@ -302,13 +318,17 @@ These are matched against the `mediaType` reported by the Confluence API. Operat
...
 
-> **Platform compatibility:** Inline images require Unique platform `2026.24.0` or later. The platform extracts searchable text from inline base64 images starting with that version; on earlier versions the page is still ingested but its embedded images contribute no searchable content. Extraction is also gated behind two company-scoped feature flags that must be enabled (`FEATURE_FLAG_ENABLE_HTML_INLINE_IMAGE_EXTRACTION_UN_20936` and `FEATURE_FLAG_ENABLE_MULTI_FILE_IMAGE_CONTENT_EXTRACTION_UN_20936`). See [Unique Platform Compatibility](./deployment.md#unique-platform-compatibility).
+Inlining is turned on by the presence of the image-extraction configuration the platform needs to extract searchable text from those inline images: set `pageIngestionConfig.htmlConfig.imageContentExtraction.languageModel` to the visual LLM configured in Knowledge Base. The connector sends this configuration with every page. When it is absent, inlining is off and referenced images fall back to standalone attachment ingestion. This replaces the previous manual step of setting the extraction config on the destination scope.
```
Ostrzy: Do you have to set both `languageModel` and `enabled: true`? Because you have that in examples but here we mention only `languageModel` 🤔

services/confluence-connector/docs/operator/deployment.md
```diff
@@ -185,7 +185,7 @@ See [Configuration -- Confluence Connection Settings](./configuration.md#Conflue
...
+Connector `2.1.0` introduces inlining of page images as base64 `data:` URIs (see [Configuration -- Image Attachments](./configuration.md#image-attachments)). Extracting searchable text from those inline images is performed platform-side and only exists from Unique `2026.24.0`. On earlier platform versions the connector still ingests the page, but embedded images contribute no searchable content. The platform-side extraction is additionally gated behind two company-scoped feature flags (`FEATURE_FLAG_ENABLE_HTML_INLINE_IMAGE_EXTRACTION_UN_20936` and `FEATURE_FLAG_ENABLE_MULTI_FILE_IMAGE_CONTENT_EXTRACTION_UN_20936`), both of which must be enabled for the company. The connector must also be configured with the extraction model via `ingestion.pageIngestionConfig` (see [Configuration -- Image Attachments](./configuration.md#image-attachments)).
```
Ostrzy: Link 🙂

services/confluence-connector/docs/technical/flows.md
```diff
@@ -207,18 +207,18 @@ An attachment is accepted if:
...
+Images embedded in a page body (via the editor's drag/drop, paste, or "Insert image" actions) are stored by Confluence as regular page attachments and surface in the same `expand=children.attachment` results. When inlining is enabled, these images are inlined directly into the page HTML (see the [Page Image Inlining](#page-image-inlining) section below) rather than ingested as separate artifacts, and image attachments are excluded from the standalone attachment path entirely. Orphan images (attached but not referenced by an in-body macro) are appended to the end of the page body. A macro-referenced image that cannot be inlined keeps its original macro and is not ingested elsewhere. Images inserted as external URLs (`<ri:url>`) are not attachments and are not ingested.
```
Ostrzy: Here we also have a wrong link 🙂

services/confluence-connector/src/synchronization/confluence-synchronization.service.ts
```diff
@@ -66,15 +67,22 @@ export class ConfluenceSynchronizationService {
         discoveredAttachments,
       );
 
+      // we either inline images in the page, either ingest them separately with no inlining.
```
Ostrzy: `We either inline images in the page, or ingest them separately with no inlining.`

Ostrzy (approved):

services/confluence-connector/docs/operator/configuration.md
lorand93: both are required.  10b97fec980c5faf017a543ba477008a6b897ce2

I updated the schema
```
 imageContentExtraction: z.object({
      enabled: z.literal(true),
      languageModel: requiredStringSchema,
    }),
    ```

PR #580, opened 2026-05-28
fix(teams-mcp): stream MS Graph recordings via Readable.fromWeb

vabatta:
## Summary

- Wrap the WHATWG `ReadableStream` returned by MS Graph SDK's `.getStream()` with `Readable.fromWeb()` for both the recording (MP4) and transcript (VTT) paths, so the subsequent `fetch()` upload to Unique storage no longer hits `TypeError: Response body object should not be disturbed or locked` on slow Graph responses. Mirrors the proven pattern in `sharepoint-connector`.
- Propagate the `Readable` type through `UniqueContentService.uploadToStorage` and `UniqueService.ingestTranscript`.

## Context

In QA we observed `TypeError: Response body object should not be disturbed or locked` during the meeting-ingestion flow on a large recording. The Graph SDK's middleware chain disturbs the WHATWG stream before it reaches `fetch(..., { body, duplex: 'half' })`. Wrapping with `Readable.fromWeb` decouples the Node stream from undici's WHATWG-stream invariants, immunizing the upload from whichever SDK middleware is to blame.

## Test plan

- [x] `pnpm -F teams-mcp check-types` — passes (only pre-existing `packages/logger` errors remain, unrelated)
- [x] `pnpm -F teams-mcp style` — passes
- [x] `pnpm -F teams-mcp build` — passes
- [ ] Manual QA in QA cluster: trigger a Teams meeting transcript+recording ingestion against a meeting large enough to cross the slow-request threshold; confirm both paths reach `Successfully completed content upload to storage`.
- [ ] Confirm no `TypeError: Response body object should not be disturbed or locked` over a 24h window of normal traffic.
- [ ] Spot-check that small/fast recordings still ingest correctly (regression smoke).

### DIFF
--- a/services/teams-mcp/src/msgraph/metrics.middleware.ts
+++ b/services/teams-mcp/src/msgraph/metrics.middleware.ts
@@ -41,40 +41,6 @@ export class MetricsMiddleware implements Middleware {
     return options?.method?.toUpperCase() || 'GET';
   }
 
-  /**
-   * Detect streaming media downloads (recording MP4, transcript VTT). Inspecting the
-   * Response body for these would risk "body disturbed/locked" errors and is meaningless
-   * for binary streams — skip the slow-request WARN that would otherwise fire for every
-   * media download since they inherently exceed the 5s threshold.
-   */
-  private isStreamingAccept(options: RequestInit | undefined): boolean {
-    const headers = options?.headers;
-    if (!headers) {
-      return false;
-    }
-
-    let accept: string | null = null;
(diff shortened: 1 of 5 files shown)


vabatta:
Closing — root cause is likely the Cilium network policy gap addressed by Unique-AG/monorepo#24704 (return packets on ephemeral ports dropped). The Readable.fromWeb wrap was defense-in-depth against a symptom; the real fix belongs in the GitOps repo for teams-mcp.

PR #616, opened 2026-06-05
fix(teams-mcp): fail loudly on non-media content downloads instead of uploading them

vabatta:
## Problem

After #614 fixed the chunked-upload rejection, recording (`video/mp4`) uploads were producing a tiny **~214-byte** blob (empty/corrupt), while transcript (`text/vtt`) uploads worked correctly.

Investigation (incl. the [Graph `callRecording` content docs](https://learn.microsoft.com/en-us/graph/api/callrecording-get?view=graph-rest-1.0&tabs=http)):

- A healthy `GET .../recordings/{id}/content` returns `200 OK` + `Content-Type: video/mp4` + the media bytes (it may 302 internally to storage; the SDK's `RedirectHandler` follows it). Our endpoint/approach is correct.
- At the SDK level, `.getStream()` and `.responseType(RAW).get()` return the **same** `rawResponse.body`, so this isn't a regression from #614 — the recording download itself is yielding ~214 bytes.
- A 214-byte payload is almost certainly a **non-media error/redirect body** — a Graph JSON error, an Azure SAS-expiry XML, or an auth/redirect HTML page — that we were uploading verbatim, because `readSizedContent` never checked `response.ok` or the content type.

## Change

`readSizedContent` now rejects non-media downloads instead of silently storing them:

- Throws when the response is **not OK** or the content type is **JSON/XML/HTML**.
- The error includes `status`, final `url`, `content-type`, `Location` header, and the (capped) response body — so the real cause is visible in logs.
- Recording upload is already wrapped in try/catch in `ingestTranscript`, so this only fails the recording (logged as a warning) and leaves transcript ingestion intact.

This is both a **correctness guard** (no more corrupt 214-byte recordings) and a **diagnostic**: the next recording attempt will surface the actual error response so we can fix the root cause (e.g. permissions, SAS expiry, or an unfollowed redirect).

Adds unit coverage for the stream / buffered / non-OK / non-media paths.

## Verification

- `tsc --noEmit`, `biome check`, and all 53 unit tests pass (4 new).

## Follow-up

Once the captured error reveals why the recording `/content` returns a non-media body, apply the targeted fix.

### DIFF
--- a/services/teams-mcp/src/utils/sized-content.spec.ts
+++ b/services/teams-mcp/src/utils/sized-content.spec.ts
@@ -1,0 +1,59 @@
+import { describe, expect, it } from 'vitest';
+import { readSizedContent } from './sized-content';
+
+async function collect(stream: ReadableStream<Uint8Array>): Promise<Uint8Array> {
+  return new Uint8Array(await new Response(stream).arrayBuffer());
+}
+
+describe('readSizedContent', () => {
+  it('streams an identity-encoded response using the advertised content-length', async () => {
+    const body = new Uint8Array(20_000).fill(7);
+    const response = new Response(body, {
+      status: 200,
+      headers: { 'content-type': 'video/mp4', 'content-length': String(body.byteLength) },
+    });
+
+    const { stream, size } = await readSizedContent(response);
+
+    expect(size).toBe(20_000);
+    expect((await collect(stream)).byteLength).toBe(20_000);
+  });
+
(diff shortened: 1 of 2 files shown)


vabatta:
Closing — going with buffering in teams-mcp (like the VTT path) instead of the non-media guard. Buffering reads the real decoded bytes and sets an exact Content-Length, which is the actual fix; superseded by the buffering PR.

PR #623, opened 2026-06-08
fix(teams-mcp): skip inaccessible recordings without dangling KB records

vabatta:
## Context

Follow-up to #619 (disk-spool upload). Investigating a participant-triggered ingest on 0.2.16 (logs provided) showed the on-demand `ingest_meeting` path **only partially failed**, and in a way that left junk behind:

- Meeting lookup, transcript metadata, recording metadata, and the **transcript upload all succeeded** for an invited attendee.
- The **recording `/content` download from Graph returned 403**: `{"statusCode":403,"code":null,"requestId":null,"body":{}}` (a `GraphError`).

Why: Teams **recordings** live in the **meeting owner's OneDrive**; `/recordings/{id}/content` 302-redirects there and is governed by that file's ACL, so a non-owner (e.g. an invited attendee triggering on-demand ingest) gets 403. **Transcripts** live in the meeting/Stream service and *are* readable by attendees. (This is orthogonal to #619 — the 403 happens on download, before any upload/spooling.)

The bug: we called `upsertContent` (which **opens storage** and returns a `writeUrl`) **before** downloading. So the 403 left a **dangling, empty content record** in the knowledge base, and the failure was swallowed as a generic warning.

## Change

**Download-first ordering** — spool the Graph download to a temp file *before* opening any content record:

- Split `uploadToStorage(writeUrl, content, mime)` into:
  - `spoolContent(content) -> { path, size, cleanup }` — streams the download to a temp file (constant memory), `fstat`s for the authoritative size; cleans up its own partial temp on failure.
  - `uploadToStorage(writeUrl, spooled, mime)` — PUTs the spooled file with an explicit `Content-Length`.
- `ingestTranscript` now does, per artifact: **spool → upsert (open) → upload → upsert (finalize)**, with `cleanup()` in `finally`.
- **Recording**: if `spoolContent` throws (e.g. 403), we **skip the recording and never open a record** — nothing to clean up — and the transcript (already ingested) stands. The 403 is detected via `error instanceof GraphError` and logged explicitly. A failure *after* the download (rare node-ingestion error) is still caught so it can't break transcript ingestion (it could leave a record, noted in a comment — teams-mcp has no delete path).
- Applied the same download-first order to the **transcript** (so a transcript download failure also leaves no dangling record).

## Why not delete-on-failure?

teams-mcp has no delete capability — platform deletes (`files.deleteByIds`) are a **GraphQL ingestion-client** operation (`@unique-ag/unique-api`), which teams-mcp doesn't wire up (it uses the public REST `FetchFn` + node-ingestion's `/scoped/upload`). Reordering **prevents** the dangling record for the 403 case instead of creating-then-deleting it — cleaner, and needs no new client/dependency.

## Limitations / follow-ups

- The recording skip is surfaced via logs/telemetry only (the tool enqueues and returns before ingestion runs, so it can't synchronously report the recording outcome).
- Letting participants actually pull the recording via **app permissions on behalf of the organizer** (`OnlineMeetingRecording.Read.All` + application access policy) is a larger follow-up for its own ticket.

## Verification

- `check-types`, `style`, `build`, `test` (49/49) all pass.
- ⏳ Manual e2e: trigger `ingest_meeting` as a non-owner attendee → transcript ingests, recording is skipped (403 logged), and **no empty recording content record** is created.

Refs: UN-20155

### DIFF
--- a/services/teams-mcp/src/unique/unique.service.ts
+++ b/services/teams-mcp/src/unique/unique.service.ts
@@ -1,4 +1,5 @@
 import { createHash } from 'node:crypto';
+import { GraphError } from '@microsoft/microsoft-graph-client';
 import { Injectable, Logger } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import { Span, TraceService } from 'nestjs-otel';
@@ -13,7 +14,7 @@ import {
   SourceOwnerType,
   UniqueIngestionMode,
 } from './unique.dtos';
-import { UniqueContentService } from './unique-content.service';
+import { type SpooledContent, UniqueContentService } from './unique-content.service';
 import { UniqueScopeService } from './unique-scope.service';
 import { UniqueUserService } from './unique-user.service';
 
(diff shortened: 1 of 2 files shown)


services/teams-mcp/src/unique/unique.service.ts
```diff
@@ -151,110 +152,150 @@ export class UniqueService {
...
+              },
+            },
+          });
+
+          span?.addEvent('recording_stored', {
+            recordingId: recording.id,
+            parentScopeId: parentScope.id,
+            childScopeId: childScope.id,
+          });
+        } catch (error) {
```
nicolae536, line 283: One thing here, if one process in this chain faile you should probably revert the others but that can be done in later pr

nicolae536 (approved):

PR #634, opened 2026-06-09
feat(mcp-server-module): un-21719 add elicitation support

nicolae536:
## Ticket
UN-21719

## Title
feat(mcp-server-module): add elicitation support to Context

## Description
- Add `elicit(schema, message)` to `Context` — accepts a Zod object schema, converts it to JSON Schema internally, and returns a fully-typed discriminated union result
- Add `elicitUrl({ elicitationId, url })` to `Context` for OAuth-style out-of-band flows, returning the user action plus a `sendCompletionNotification()` callback
- Both methods throw `McpError(InternalError)` in stateless mode where no persistent client connection exists
- Export `FormElicitResult<T>` and `UrlElicitResult` types from the package
- Document both methods in README with usage examples

### DIFF
--- a/packages/mcp-server-module/src/services/handlers/mcp-handler.base.ts
+++ b/packages/mcp-server-module/src/services/handlers/mcp-handler.base.ts
@@ -124,14 +124,14 @@ export abstract class McpHandlerBase {
         message: string;
         url: string;
       }): Promise<UrlElicitResult> => {
+        const sendCompletionNotification =
+          mcpServer.server.createElicitationCompletionNotifier(elicitationId);
         const result = await mcpServer.server.elicitInput({
           mode: 'url',
           elicitationId,
           message,
           url,
         });
-        const sendCompletionNotification =
-          mcpServer.server.createElicitationCompletionNotifier(elicitationId);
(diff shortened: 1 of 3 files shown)


packages/mcp-server-module/src/services/handlers/mcp-handler.base.ts
```diff
@@ -63,6 +75,38 @@ export abstract class McpHandlerBase {
...
+        message: string,
+      ): Promise<FormElicitResult<T>> => {
+        const result = await mcpServer.server.elicitInput({
+          message,
+          requestedSchema: z.toJSONSchema(schema, {
+            io: 'input',
+          }) as ElicitRequestFormParams['requestedSchema'],
+        });
+        // SDK validates content against the JSON Schema derived from schema; cast is safe.
+        return result as FormElicitResult<T>;
```
vabatta: Can you bully and see if the assertion is necessary? Like, isn't the underlaying SDK typed properly? Why we need `as ...`?

nicolae536: Fixed

vabatta (approved):

PR #653, opened 2026-06-22
feat(teams-mcp): interactive disambiguation + richer list metadata

vabatta:
## Context

When a user asks the model to act on a Teams chat/team/channel by name and two entities share that name, resolution threw a `ConflictException` ("matches multiple… Please be more specific.") with **no way for the model to disambiguate** — a dead end. The codebase already anticipated the fix with TODOs pointing at `context.elicitInput()` "once exposed on the Context interface." That capability is now exposed (`Context.elicit(schema, message)`), so this PR wires it up.

## What changed

1. **Interactive picker.** New shared `disambiguate()` helper (`chat/disambiguate.ts`) presents a `context.elicit()` single-select when a name matches multiple entities. Labels are index-prefixed (`1.`, `2.`) so identical labels stay distinct.
2. **Graceful fallback.** Falls back to today's `ConflictException` when: >10 matches (a 50-option picker is worse than asking for specificity), the user declines/cancels, or `elicit` throws (stateless transport / client lacks the capability). No regression for non-elicitation clients.
3. **Applies to chats, teams, and channels.** `Context` is threaded through the three resolvers and the tools that call them; both stale TODOs removed.
4. **Richer list metadata** (so picker labels are meaningful and the model has context up front):
   - `list_chats`: `createdDateTime` + `lastMessageAt` (via `$expand=lastMessagePreview`)
   - `list_channels`: `createdDateTime` + `membershipType`
   - `list_teams`: `isArchived`
   - Tool descriptions updated to advertise the new metadata.

### Note on team metadata
The plan called for adding `visibility` and `webUrl` to teams. Live Graph docs (`user-list-joinedteams`) confirm `/me/joinedTeams` returns those — plus `createdDateTime` — as **always `null`**; only `id`, `displayName`, `description`, `isArchived`, and `tenantId` are populated. Shipping always-null fields would mislead the model, so I added `isArchived` (the one useful populated field for telling same-named teams apart) instead.

## Verification
- `pnpm run check-types` → clean
- `pnpm build` → success
- `biome check` → clean

## Out of scope
Exposing raw IDs as tool params (rejected in favor of the picker), URL-mode elicitation, server composition, pagination redesign. No unit tests — this MCP has no test suite yet and adding one is out of scope for this PR.

### DIFF
--- a/services/teams-mcp/src/chat/disambiguate.ts
+++ b/services/teams-mcp/src/chat/disambiguate.ts
@@ -1,7 +1,9 @@
 import assert from 'node:assert';
 import { type Context } from '@unique-ag/mcp-server-module';
-import { ConflictException } from '@nestjs/common';
+import { ConflictException, Logger } from '@nestjs/common';
 import * as z from 'zod';
+
+const logger = new Logger('disambiguate');
 
 // Above this many matches a single-select picker is more hindrance than help;
 // fall back to asking the user to be more specific instead.
@@ -63,7 +65,13 @@ export async function disambiguate<T>(
       throw err;
     }
     // Otherwise elicit itself threw — stateless mode or the client lacks the
-    // elicitation capability. Fall back to today's behaviour.
+    // elicitation capability. Log it so a genuinely unexpected error (rather
+    // than the expected "elicitation unsupported") is still visible, then fall
+    // back to today's behaviour.
+    logger.warn(
+      { err: err instanceof Error ? err.message : String(err) },
+      'Elicitation failed; falling back to ConflictException',
+    );
     throw new ConflictException(opts.conflictMessage);
   }
 }
--- a/services/teams-mcp/src/chat/disambiguate.ts
+++ b/services/teams-mcp/src/chat/disambiguate.ts
@@ -1,0 +1,69 @@
+import assert from 'node:assert';
+import { type Context } from '@unique-ag/mcp-server-module';
+import { ConflictException } from '@nestjs/common';
+import * as z from 'zod';
+
+// Above this many matches a single-select picker is more hindrance than help;
+// fall back to asking the user to be more specific instead.
+const PICKER_MAX = 10;
+
+/**
+ * Resolve an ambiguous set of candidate entities to a single one by presenting
+ * the user an interactive single-select picker via `context.elicit()`.
+ *
+ * Falls back to throwing the original `ConflictException` when the picker is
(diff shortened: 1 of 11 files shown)


services/teams-mcp/src/chat/disambiguate.ts
```diff
@@ -0,0 +1,69 @@
...
+    assert.ok(picked, 'elicited selection did not match any candidate');
+    return picked.item;
+  } catch (err) {
+    // A decline/cancel surfaces as the ConflictException thrown above — rethrow it.
+    if (err instanceof ConflictException) {
+      throw err;
+    }
+    // Otherwise elicit itself threw — stateless mode or the client lacks the
+    // elicitation capability. Fall back to today's behaviour.
+    throw new ConflictException(opts.conflictMessage);
```
Ostrzy, line 75: A bit of a nitpick, but maybe we should log something here, so we're sure we can see in logs when we wrap error that is not necessarily from elicit itself? 🙂

Ostrzy (approved):

services/teams-mcp/src/chat/disambiguate.ts
vabatta: Good call — agreed. The broad catch is deliberate (the "elicitation unsupported" `McpError` isn't reliably type-detectable), but that means any unexpected throw in the try would get silently rewritten into the "be more specific" `ConflictException`. Added a warn-level log of the wrapped error before the fallback in d579eeed so it stays visible in logs.

services/teams-mcp/src/chat/disambiguate.ts
```diff
@@ -0,0 +1,77 @@
...
+      throw err;
+    }
+    // Otherwise elicit itself threw — stateless mode or the client lacks the
+    // elicitation capability. Log it so a genuinely unexpected error (rather
+    // than the expected "elicitation unsupported") is still visible, then fall
+    // back to today's behaviour.
+    logger.warn(
+      { err: err instanceof Error ? err.message : String(err) },
+      'Elicitation failed; falling back to ConflictException',
+    );
```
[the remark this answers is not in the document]
vabatta: Agreed — and it reinforces the previous comment's intent: the log exists to spot real defects among the expected unsupported-elicitation cases, and the stack trace is what tells them apart. Now passing the full error object (`{ err }`) so the logger's error serializer keeps the stack and type. Fixed in 298a8610.

PR #661, opened 2026-06-25
fix(outlook-semantic-mcp): un-22182 fix delegated mailboxes web links

nicolae536:
## Problem

Since November 2025 Microsoft has been migrating tenants to the new `outlook.cloud.microsoft` domain. After migration, every Graph API endpoint returns `webLink` values in a new format that embeds a **RestImmutableEntryId** (`AAkA…`). The classic OWA deep-link format — the only reliable way to open a message in a delegated or shared mailbox — requires a **RestId** (`AAMk…`) in its `ItemID` query parameter. Passing an immutable ID there produces a broken link that OWA cannot resolve.

Additionally, reply drafts for emails with immutable IDs links were broken.

## Fix

- **`BuildWebLinksCommand`** (new): for own-mailbox items the stored `webLink` is returned as-is (both OWA formats work for the authenticated user). For delegated/shared mailboxes a classic OWA URL is always constructed — immutable IDs are first translated to RestIds via the Graph `translateExchangeIds` API, then the URL is built with the RestId.
- **KQL search**: the `$search` endpoint always returns webLinks in the classic OWA format with a RestId, so delegated-mailbox results now expose the actual link instead of suppressing it.
- **Semantic search**: replaced the blanket link suppression for delegated mailboxes with a call to `BuildWebLinksCommand`, which constructs a working OWA URL for every result regardless of mailbox ownership.
- **Reply drafts**: added `idIsImmutable` plumbing through the reply path; when true, the `Prefer: IdType="ImmutableId"` header is set on the `createReplyAll` call. Reply body is now sent as `{ comment: htmlContent }` to preserve the quoted thread. The created draft's link is also built via `BuildWebLinksCommand`.
- **`replyToParams`**: search results now expose `isReplyable: false` for draft messages so the LLM does not attempt to reply to a draft.

---

- Fixes [UN-22182](https://unique-ch.atlassian.net/browse/UN-22182)
- Fixes web links in general for shared mailboxes

Refs: UN-22182

[UN-22182]: https://unique-ch.atlassian.net/browse/UN-22182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


services/outlook-semantic-mcp/src/features/content/search/semantic-search-emails.query.ts
```diff
@@ -172,7 +177,7 @@ export class SemanticSearchEmailsQuery {
       });
     }
 
-    const results = pipe(
+    const rawResults = pipe(
```
vabatta, line 180: Interesting mixture of `functional` in `imperative` code, but looks good :)
If only we were using an _effect_ive framework for it...

vabatta (approved):

PR #663, opened 2026-06-29
fix(teams-mcp): attribute upstream Microsoft/network failures to the consumer

vabatta:
## Why

Production logs for `teams-mcp` (one pod, 17-min window on 2026-06-26) showed 15 errors that were **not** our product's fault but reached the MCP consumer as opaque internal errors (a raw `error.message` like `"Error while processing response."`):

- **8×** genuine Microsoft Graph `HTTP 500 InternalServerError` on `GET /chats/{id}/messages` — deterministic per-chat (a Graph-side fault on specific chat content).
- **6×** DNS/network failures (`fetch failed` / `EAI_AGAIN` to graph.microsoft.com).
- **1×** DNS failure to the Postgres cluster.

This made our product look broken when the third party was at fault. This PR surfaces upstream faults **loudly and attributes them honestly**, while staying neutral on genuine `4xx` (our request / the user's input) so attribution stays credible.

## What

**1. Error classifier — `src/utils/classify-error.ts`** (pure functions)
- `isUpstreamNetworkError(error)` — true for `fetch failed` or a transport errno (`EAI_AGAIN`, `ENOTFOUND`, `ECONNRESET`, … `UND_ERR_CONNECT_TIMEOUT`, `UND_ERR_SOCKET`) at the error or `error.cause` level.
- `classifyError(error)` → `{ fault, httpStatus?, graphCode?, requestId?, retryable, message }`:
  - GraphError `429` / `5xx` → `upstream_microsoft` (names Microsoft, includes HTTP status + Graph `code` + `requestId` for MS support).
  - GraphError `401` → `auth`; GraphError `4xx` → `client_request` (neutral, **no Microsoft blame**).
  - network → `upstream_network`; else → `unknown`.
- `toAttributedError(error)` — `McpError` unchanged (preserves the reauth flow), `unknown`-fault unchanged (no masking real bugs), otherwise a new `Error` with the attributed message.

**2. Tool boundary — `@AttributeUpstreamErrors()` decorator** applied to all 14 tools, **above `@Span()`** so the span still records the *original* error before attribution rewrites it. The MCP module wraps a thrown `Error` into `{ content, isError: true }`, so throwing the attributed message is sufficient and keeps each tool's typed `outputSchema` intact.

**3. Metrics one-line fix — `src/msgraph/metrics.middleware.ts`** catch block now sets `status_class: 'network'` for transport failures instead of hardcoding `'5xx'`, so per-status-class queries separate cleanly (`5xx` = Microsoft server fault, `network` = connectivity/DNS, `4xx` = our/user request).

### Why no shared-module / interceptor seam
Tool execution runs inside the MCP SDK transport, **outside** the NestJS pipeline, so `APP_INTERCEPTOR` / `APP_FILTER` (incl. the existing `GraphErrorFilter`) never see tool errors. The fix is contained in teams-mcp at the consumer-facing edge.

## Out of scope
- Retry-policy widening / "graceful degradation" — a true 500 returns no body to salvage; attribution is the fix.
- Shared `mcp-server-module` `mapError` hook + structured `_meta` — the central alternative, not chosen.

## Testing
- `pnpm --filter teams-mcp test` → 67/67 pass (new specs for the classifier and the metrics `network` vs `5xx` split).
- `tsc --noEmit` and `biome check` clean.
- **Not done (needs a live instance + Microsoft creds):** end-to-end smoke against an invalid `chatId` (expect neutral `client_request` message) and a known failing chat id (expect the loud Microsoft-500 attribution).

### DIFF
--- a/services/teams-mcp/src/msgraph/metrics.middleware.ts
+++ b/services/teams-mcp/src/msgraph/metrics.middleware.ts
@@ -3,6 +3,7 @@ import { Logger } from '@nestjs/common';
 import type { Counter, Histogram } from '@opentelemetry/api';
 import { MetricService } from 'nestjs-otel';
 import { serializeError } from 'serialize-error-cjs';
+import { isUpstreamNetworkError } from '../utils/classify-error';
 import { normalizeError } from '../utils/normalize-error';
 
 export class MetricsMiddleware implements Middleware {
@@ -157,7 +158,8 @@ export class MetricsMiddleware implements Middleware {
       this.msgraphRequestCounter.add(1, {
         endpoint,
         method,
-        status_class: '5xx',
+        // Distinguish "couldn't reach Microsoft" (DNS/connectivity) from "Microsoft's server errored".
+        status_class: isUpstreamNetworkError(error) ? 'network' : '5xx',
       });
 
       this.msgraphRequestDuration.record(duration / 1000, {
--- a/services/teams-mcp/src/utils/classify-error.ts
+++ b/services/teams-mcp/src/utils/classify-error.ts
@@ -1,0 +1,165 @@
+import { GraphError } from '@microsoft/microsoft-graph-client';
+import { McpError } from '@modelcontextprotocol/sdk/types.js';
+
+/**
+ * Transport-level error codes that mean we could not reach Microsoft Graph at all
+ * (DNS, connection, socket failures) — as opposed to Microsoft returning an HTTP error.
+ *
+ * Mirrors the retryable set used by the SharePoint connector's undici dispatcher
+ * (`sharepoint-rest-http.service.ts`), plus `EAI_AGAIN` / `UND_ERR_CONNECT_TIMEOUT`
+ * observed in teams-mcp production logs.
+ */
+const UPSTREAM_NETWORK_CODES = new Set([
+  'EAI_AGAIN',
+  'ENOTFOUND',
+  'ECONNRESET',
+  'ECONNREFUSED',
+  'ETIMEDOUT',
+  'ENETUNREACH',
+  'ENETDOWN',
+  'EHOSTUNREACH',
+  'EPIPE',
+  'UND_ERR_CONNECT_TIMEOUT',
+  'UND_ERR_SOCKET',
+]);
+
+/**
+ * The party at fault for a failed operation.
+ * - `upstream_microsoft` — Microsoft Graph returned 429 or 5xx (their server erred).
+ * - `upstream_network`    — we could not reach Microsoft Graph (connectivity / DNS).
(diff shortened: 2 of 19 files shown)


services/teams-mcp/src/utils/classify-error.ts
```diff
@@ -0,0 +1,165 @@
...
+    const cause = errno.cause as NodeJS.ErrnoException | undefined;
+    const code = errno.code ?? cause?.code;
+    return {
+      fault: 'upstream_network',
+      retryable: true,
+      message:
+        `Could not reach Microsoft Graph` +
+        (code ? ` (network error: ${code})` : ` (network error)`) +
+        `. This is a connectivity or DNS problem between this connector and Microsoft, not a defect in your request or this connector. This is usually transient — please retry shortly.`,
+    };
```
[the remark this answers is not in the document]
vabatta: Good catch — fixed in 6696f01b.

The tool boundary catches errors from *every* dependency, so the same transport errnos (`EAI_AGAIN`, `fetch failed`, …) from Postgres/Drizzle or the Unique API were being reported as "Could not reach Microsoft Graph". (The original incident even included a DNS failure to the Postgres cluster, so this was a real false-attribution path.)

`classifyError` now only returns `upstream_network` when the unreachable host is a confirmed Microsoft endpoint (`graph.microsoft.com` / `login.microsoftonline.com` + national clouds), read from the error/cause `.hostname` that Node sets on DNS failures. A non-Microsoft host (or a hostless transport error we can't confidently attribute) falls through to `unknown` and is passed through unchanged rather than blamed on Graph.

`isUpstreamNetworkError` stays a host-agnostic transport predicate — that's intentional and correct for its only other caller, the Graph metrics middleware, which is Graph-scoped by construction. Added tests for the Microsoft-host, non-Microsoft-host, and hostless cases.

services/teams-mcp/src/msgraph/metrics.middleware.ts
```diff
@@ -157,7 +158,8 @@ export class MetricsMiddleware implements Middleware {
       this.msgraphRequestCounter.add(1, {
         endpoint,
         method,
-        status_class: '5xx',
+        // Distinguish "couldn't reach Microsoft" (DNS/connectivity) from "Microsoft's server errored".
+        status_class: isUpstreamNetworkError(error) ? 'network' : '5xx',
```
irina-unique, line 162: shall we say 'upstream-network'?

irina-unique (approved):

services/teams-mcp/src/msgraph/metrics.middleware.ts
vabatta: `upstream` is the semantics for "us" as consumer-facing hence `network` is semantically _more_ correct.

PR #665, opened 2026-06-29
fix(confluence-connector): make mermaid diagrams publishable

lorand93:
## Why
Fix https://github.com/Unique-AG/connectors/actions/runs/28379691066/job/84078639538 

The `[confluence-connector] Publish Docs to Confluence` workflow was failing: `md2conf` renders Mermaid diagrams with `--render-mermaid`, and Mermaid's parser cannot handle raw `<`, `>`, or `&` inside sequence-diagram note/message labels. The `flows.md` content-sync diagram referenced `<ac:image>`, ``, and `?spaceKey=&title=`, so the publish job aborted on `flows.md` and never updated Confluence (and therefore docs.unique.ai).

## What

Escape those characters using Mermaid entity codes so the diagrams parse **and** keep their exact technical notation:

- `#60;` renders as `<`
- `#62;` renders as `>`
- `#38;` renders as `&`

This preserves the real `<ac:image>` macro, the `` element, and the `?spaceKey=&title=` query string in the rendered diagrams, rather than dropping the notation.

## Verification

Rendered all 16 Mermaid diagrams under `services/confluence-connector/docs/` locally with `@mermaid-js/mermaid-cli` — 0 failures, and the output shows the angle brackets and ampersand correctly.

### DIFF
--- a/services/confluence-connector/docs/README.md
+++ b/services/confluence-connector/docs/README.md
@@ -189,7 +189,7 @@ sequenceDiagram
 
         Connector->>Confluence: Fetch full page content
         Confluence->>Connector: Page HTML body
-        opt Page has ac:image references to Confluence attachments
+        opt Page has #60;ac:image#62; references to Confluence attachments
             opt Some references point to another page
                 Connector->>Confluence: Look up target page by (spaceKey, title)
                 Confluence->>Connector: Target page + its attachments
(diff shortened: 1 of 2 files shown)


florianrueegg (approved):

lorand93:
bugbot review

PR #691, opened 2026-07-14
feat(sharepoint-connector): replace noisy sync failure alert with SyncDegraded

unique-adrian:
## Summary
- Replace the noisy `SharepointConnectorSyncFailures` default alert with `SharepointConnectorSyncDegraded` in the chart's `prometheus.additionalAlerts`.
- The new alert scopes to site syncs (`sync_type="site"`), evaluates the failure ratio over a ~5-cycle `75m` window, and fires only above a `0.5` ratio sustained for `15m` (vs. the old `5m` window, `>0.01` ratio, `30s` for).
- This removes false positives caused by transient, self-healing sync failures while still catching sustained degradation.

## Details
- `values.yaml`: swapped the alert definition (expr, `for`, threshold, summary/description, trimmed runbook to the 3 relevant steps).
- `README.md`: regenerated (hand-updated) the helm-docs value rows for the renamed alert.
- `docs/operator/configuration.md`: updated the default-alerts table row.

The chart `version`/`appVersion` bump is handled automatically by release-please from this `feat` commit; no manual version edit is included.

## Notes / open question
- The alert was **renamed** (`SharepointConnectorSyncFailures` → `SharepointConnectorSyncDegraded`), so any external dashboards, silences, or routing referencing the old name must be updated. If preferred, I can keep the old name to avoid that churn.

## Test plan
- [ ] `helm template` renders the `SharepointConnectorSyncDegraded` PrometheusRule with the expected expr/labels.
- [ ] Confirm the PromQL expression is valid against the `spc_sync_duration_seconds_count` series with `sync_type="site"`.
- [ ] Verify release-please opens a version-bump PR after merge.

### DIFF
--- a/services/sharepoint-connector/deploy/helm-charts/sharepoint-connector/README.md
+++ b/services/sharepoint-connector/deploy/helm-charts/sharepoint-connector/README.md
@@ -142,14 +142,14 @@ spec:
 | prometheus.additionalAlerts.SharepointConnectorGraphQLErrors.for | string | `"30s"` |  |
 | prometheus.additionalAlerts.SharepointConnectorGraphQLErrors.labels.alertGroup | string | `"sharepoint-connector"` |  |
 | prometheus.additionalAlerts.SharepointConnectorGraphQLErrors.labels.severity | string | `"warning"` |  |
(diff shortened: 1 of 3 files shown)


lorand93 (approved):
looks good

PR #693, opened 2026-07-16
feat(teams-mcp): ensure service-user permissions on root scope at startup

vabatta:
## Context

`teams-mcp` ingests Teams transcripts/recordings by creating sub-scopes under a pre-provisioned **root scope** (`UNIQUE_ROOT_SCOPE_ID`) and granting access on those sub-scopes. It **assumes** the service user already has `MANAGE`/`WRITE`/`READ` on the root scope — but nothing establishes or verifies that. If the service user lacks those accesses, `createScope`/`add-access` under the root fails at ingestion time with an opaque platform error.

`sharepoint-connector` solves the equivalent problem with `InitializeRootScopeCommand`. This ports that idea to teams-mcp as a **startup bootstrap** so misconfiguration surfaces immediately at boot instead of mid-ingestion.

## Changes

- **New `RootScopeBootstrapService`** (`src/unique/root-scope-bootstrap.service.ts`) implementing `OnApplicationBootstrap`. It reads `unique.rootScopeId` and `serviceExtraHeaders['x-user-id']` from config and self-grants the service user `MANAGE`/`READ`/`WRITE` on the root scope via the existing `UniqueScopeService.addScopeAccesses`. On failure it logs a clear error and **rethrows** so the pod hard-fails at boot (Nest rejects `NestFactory.create` when `onApplicationBootstrap` throws).
- **Wired into `UniqueModule`** providers (not exported; Nest runs the lifecycle hook automatically).
- **Unit tests** covering the happy path (exactly three USER accesses on the scope), hard-fail on grant rejection, and the missing-`x-user-id` assertion guard.

teams-mcp is simpler than sharepoint here: the service user id is already known (the `x-user-id` header, required in both auth modes) and the root scope is pre-provisioned — so no `me` lookup, no scope creation, no external-id claiming.

## Out of scope

- No new env var, no `me`/getCurrentUser endpoint, no scope existence pre-check (a bad `rootScopeId` already hard-fails via `add-access`).
- No per-ingestion re-verification (startup-only).

## Verification

- Unit: `pnpm --filter teams-mcp test` (3/3 pass)
- Typecheck: `pnpm --filter teams-mcp build`
- Lint: Biome clean

### DIFF
--- a/services/teams-mcp/docs/operator/configuration.md
+++ b/services/teams-mcp/docs/operator/configuration.md
@@ -250,7 +250,8 @@ The root scope must be created **manually** in the Unique platform before deploy
 
 3. **Grant the service account access**
 
-   - Ensure the Zitadel service account (see above) has read/write permissions on the root scope
+   - Ensure the Zitadel service account (see above) has `MANAGE`, `READ`, and `WRITE` permissions on the root scope
+   - On startup the server self-grants these three access types to the service user (the `x-user-id` from the API headers) on the configured root scope, so a correctly configured deployment converges automatically. If that grant fails (e.g. the scope ID is wrong or the service account lacks permission to manage it), the pod **hard-fails at boot** rather than surfacing the error mid-ingestion — check the startup logs for `root scope <id> permission bootstrap failed`.
 
 4. **Configure in Helm Values**
    ```yaml
--- a/services/teams-mcp/docs/operator/configuration.md
+++ b/services/teams-mcp/docs/operator/configuration.md
@@ -250,8 +250,8 @@ The root scope must be created **manually** in the Unique platform before deploy
 
 3. **Grant the service account access**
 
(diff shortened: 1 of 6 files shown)


lorand93:
Grok 4.5's findings:
 > I’d not ship as-is until the docs mismatch and the e2e/init side-effect are addressed, and until someone confirms Public API folder/add-access can actually grant when the service user has no prior access on that scope (sharepoint’s GraphQL path explicitly can).
 
image

 Basically he flags that teams mcp took the public route and wants to make sure that it works 
 > confirm Public API can self-grant from zero
 
image

 Docs must also be updated: 
> docs still describe the old manual model
docs/operator/configuration.md still says operators must manually grant read/write on the root. The PR auto-grants MANAGE/READ/WRITE and hard-fails boot on failure. Operators will be misled (Bugbot’s low-severity note is underplaying this).

lorand93 (approved):

PR #697, opened 2026-07-22
feat(teams-mcp): make Unique integration optional via UNIQUE_INTEGRATION

nicolae536:
## Summary
- Add `UNIQUE_INTEGRATION=enabled|disabled` so Unique/Zitadel config (including `rootScopeId`) is required only when enabled
- KB tools (`start_kb_integration`, `stop_kb_integration`, `verify_kb_integration_status`, `ingest_meeting`) fail fast when disabled; services assert before using Unique config
- Helm/docs/.env.example updated; chart emits other `UNIQUE_*` only when integration is enabled

## Test plan
- [ ] Boot with `UNIQUE_INTEGRATION=disabled` and no other Unique env vars — app starts, chat tools work
- [ ] Call KB tools while disabled — each returns a clear disabled/misconfigured error
- [ ] Boot with `UNIQUE_INTEGRATION=enabled` and full Unique config — existing ingestion/KB flows still work
- [ ] Boot with `UNIQUE_INTEGRATION=enabled` but missing `UNIQUE_ROOT_SCOPE_ID` — startup config validation fails
- [ ] Helm render with `integration: disabled` — only `UNIQUE_INTEGRATION` emitted among Unique vars


services/teams-mcp/src/config/unique.config.ts
```diff
@@ -73,13 +61,40 @@ const baseConfig = z.object({
...
+  integration: z
+    .literal('disabled')
+    .describe('Unique integration is disabled; chat-only mode without knowledge base features.'),
 });
 
-export const UniqueConfigSchema = z
-  .discriminatedUnion('serviceAuthMode', [clusterLocalConfig, externalConfig])
-  .and(baseConfig);
+// Enabled variants are separate object schemas (Zod 4 discriminatedUnion does not support `.and()`).
+export const UniqueConfigSchema = z.union([
```
Ostrzy: Probably should be discriminatedUnion

services/teams-mcp/src/transcript/tools/ingest-meeting.tool.ts
```diff
@@ -84,6 +91,15 @@ export class IngestMeetingTool {
...
     }
 
+    if (!isUniqueIntegrationEnabled(this.config.get('unique', { infer: true }))) {
+      return {
+        success: false,
+        message: UNIQUE_INTEGRATION_DISABLED_TOOL_MESSAGE,
+        meeting: null,
+        queued: [],
+      };
+    }
```
Ostrzy: As we discussed, we should stop exposing ingestion-related tools when the unique integration is disabled

services/teams-mcp/src/unique/unique.module.ts
```diff
@@ -25,6 +29,13 @@ import { UniqueUserMappingService } from './unique-user-mapping.service';
       inject: [ConfigService],
       useFactory(config: ConfigService<UniqueConfigNamespaced, true>): FetchFn {
         const uniqueConfig = config.get('unique', { infer: true });
+        if (uniqueConfig.integration === 'disabled') {
+          return (async () => {
+            throw new Error(UNIQUE_INTEGRATION_MISCONFIGURED_MESSAGE);
+          }) as FetchFn;
+        }
+
+        assertUniqueIntegrationEnabled(uniqueConfig);
```
Ostrzy: Do we need this assert with the above funniness?

services/teams-mcp/src/unique/unique.module.ts
```diff
@@ -25,6 +29,13 @@ import { UniqueUserMappingService } from './unique-user-mapping.service';
       inject: [ConfigService],
       useFactory(config: ConfigService<UniqueConfigNamespaced, true>): FetchFn {
         const uniqueConfig = config.get('unique', { infer: true });
+        if (uniqueConfig.integration === 'disabled') {
+          return (async () => {
+            throw new Error(UNIQUE_INTEGRATION_MISCONFIGURED_MESSAGE);
+          }) as FetchFn;
+        }
```
Ostrzy: As we discussed, we should make an injection token for enabled config so it can be used in all the places that expect enabled unique config. Then we could get rid of all the asserts.

Ostrzy (dismissed):

services/teams-mcp/src/config/unique.config.ts
nicolae536: Done

services/teams-mcp/src/transcript/tools/ingest-meeting.tool.ts
nicolae536: Done

services/teams-mcp/src/unique/unique.module.ts
nicolae536: Done

services/teams-mcp/src/unique/unique.module.ts
nicolae536: Done

services/teams-mcp/src/config/unique.config.ts
```diff
@@ -56,30 +77,25 @@ const externalConfig = z.object({
...
+    .describe('Unique integration is disabled; chat-only mode without knowledge base features.'),
 });
 
-export const UniqueConfigSchema = z
-  .discriminatedUnion('serviceAuthMode', [clusterLocalConfig, externalConfig])
-  .and(baseConfig);
+export const UniqueConfigSchema = z.discriminatedUnion('integration', [
+  enabledUniqueConfig,
+  disabledUniqueConfig,
+]);
```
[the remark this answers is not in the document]
nicolae536: Verified empirically against zod `4.1.5` (the `zod/v4` entrypoint this codebase uses) — the nesting works correctly here, this doesn't apply.

Zod v3's `discriminatedUnion` required every option to be a plain `ZodObject` with a `.shape`, which is what broke nesting a discriminated union inside another one. Zod v4's `$ZodDiscriminatedUnion` builds its dispatch table from each option's `propValues` (the set of literal values each field can take), computed recursively — a nested discriminated union merges the `propValues` of its own options, so the outer union correctly resolves `integration` through the inner `serviceAuthMode` union.

Ran the real `UniqueConfigSchema` from this branch through 5 cases:
- `{ integration: 'disabled' }` → parses
- `{ integration: 'enabled', serviceAuthMode: 'external', ... }` → parses, narrowed correctly
- `{ integration: 'enabled', serviceAuthMode: 'cluster_local', ... }` → parses, narrowed correctly
- enabled config missing `rootScopeId` → rejected with a clear `invalid_type` error
- garbage `integration` value → rejected with `invalid_union` / "No matching discriminator"

Schema construction and parsing both behave as expected, so leaving the nested `discriminatedUnion` as-is.

florianrueegg (approved):

PR #712, opened 2026-07-28
feat(hello-mcp,ci,scripts,main): add Python hello-mcp FastMCP under services/

proschu2:
## Summary
- Add a minimal no-auth FastMCP Hello World under `services/hello-mcp` (tool `hello`, `GET /probe`, HTTP MCP on `/mcp`, port 8000).
- Ship Dockerfile (digest-pinned bases) + lean Helm chart (Unique `base` + Kong ingress stub).
- Keep Python out of pnpm/turbo (`pnpm-workspace` exclude + no `package.json`); Node `services-ci` only matrices dirs with `package.json`; Python work runs via `python-ci.yaml`.
- Wire release-please / manual dispatch using the standard `services/<name>` CD paths.

## Test plan
- [ ] `cd services/hello-mcp && uv sync && uv run pytest`
- [ ] `uv run hello-mcp` then `curl localhost:8000/probe`
- [ ] Docker build with repo-root context + Dockerfile under `services/hello-mcp/deploy`
- [ ] `helm template` with `values.yaml` + `ci-values.yaml`
- [ ] Confirm `pnpm -r list` does not include hello-mcp
- [ ] Confirm PR CI runs Python jobs for hello-mcp and does not run Node template CI on it


konsti:
I would not add a `python` folder, but rather mix the Python services under services and configure turborepo and pnpm to ignore the service.

proschu2:
Addressed review feedback:
- Moved `hello-mcp` to `services/hello-mcp` (removed `python/` top-level folder)
- Excluded it from pnpm via `pnpm-workspace.yaml` (`!services/hello-mcp`); no `package.json` so turbo ignores it too
- Node `services-ci` only matrices services with `package.json`; Python CI covers `pyproject.toml` services
- Reverted `_template-cd` `service-path` (standard `services/<name>` paths again)
- Pinned Dockerfile base images with digests (Bugbot)

services/hello-mcp/deploy/helm-charts/hello-mcp/values.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+rabbitmq:
+  enabled: false
+
+service:
+  port: 80
+
+# Prefer custom ingress.yaml over base HTTPRoute for MCP hostnames.
+routes:
+  hostname: ""
```
pwoelfle-unique: Why using Kubernetes Ingress instead of Kubernetes Gateway API?

pwoelfle-unique (approved):

services/hello-mcp/deploy/helm-charts/hello-mcp/values.yaml
```diff
@@ -0,0 +1,80 @@
...
+  ingressClassName: kong
+  hosts: []
+  tls: []
+
+hooks:
+  migration:
+    enabled: false
+
+networkPolicy:
+  enabled: false
```
pwoelfle-unique, line 106: You must place the network policy configuration from https://github.com/Unique-AG/monorepo/pull/27491 `gitops-resources/argocd/clusters/unique/qa/application-specs/mcp/hello/network-policy.values.yaml` here and keep `enabled: false`-  see also https://github.com/Unique-AG/monorepo/blob/master/next/apps/chat/deploy/helm-chart/values.yaml#L182-L191

The kong ingress rule is automatically created when specifying the internalServices.dependents block like here https://github.com/Unique-AG/monorepo/blob/master/next/apps/chat/deploy/helm-chart/values.yaml#L134-L144

pwoelfle-unique (changes requested):

nicolae536 (changes requested):
There are a few things which I would like to know if they are already configured cause in the base setup they should be

1. Logs formatting - only if possible we should try to standardise the logs formatting between the python / typescript mcp's our colleague @Ostrzy started a base setup for Edgar mcp but it was never merged since the project was postponed [link](https://github.com/Unique-AG/connectors/pull/714/changes)

2. Env vars => on the typescript side we can easily define objects with the env vars and they are validated so that the mcp startup fails

3. Lint in the ci to keep code consistent - can we configure python to have explicit return types and type definitions, this would help during code reviews, maybe I'm from a different world but from what I read you could configure this in python

4. Maybe we can also configure db / rabbitmq - just to have it in the template and remove it per project basis

Please take a look at the edgar pr and let's discuss what it makes sense and what not.

services/hello-mcp/deploy/helm-charts/hello-mcp/values.yaml
proschu2: For hello we keep both off by default in the chart:
- `routes` (Gateway API / HTTPRoute) — preferred for Unique when hostname is a single fixed host (no MCP wildcard need)
- `ingress` — still available for customers / wildcard cases like Teams/Outlook

Unique QA overlay can enable `routes` + hostname; chart stays portable. Added `internalServices.dependents.ingressGateway` + NetworkPolicy stub with `enabled: false` (shape aligned with monorepo mcp/hello NP).

services/hello-mcp/deploy/helm-charts/hello-mcp/values.yaml
proschu2: Done in the chart `values.yaml`: NetworkPolicy stub (Cilium, host+Kong on 8000, baseline prometheus/ephemeralPorts off) with `enabled: false`, plus `internalServices.dependents.ingressGateway` so Unique overlays can flip NP on without reinventing Kong access.

services/hello-mcp/deploy/helm-charts/hello-mcp/values.yaml
proschu2: Follow-up: we dropped the custom Ingress template entirely. Exposure is Gateway API only (`routes` via the `base` chart). Chart defaults leave `routes.hostname` empty (no HTTPRoute until the Unique overlay sets hostname + `paths.*.enabled`). NP stub (`enabled: false`) + `internalServices.dependents.ingressGateway` remain as discussed.

proschu2:
@nicolae536 
1. we can discuss it - we'd need to define the format in `unique_mcp`
2. yes, doable in py. Env vars are needed e.g. for URL definition and ZITADEL setup
3. added
4. there with enabled: false

services/hello-mcp/deploy/helm-charts/hello-mcp/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+              protocol: TCP
+    - fromEndpoints:
+        - matchLabels:
+            "k8s:io.kubernetes.pod.namespace": system
+            "k8s:app.kubernetes.io/name": gateway
+            "k8s:app.kubernetes.io/instance": kong
+      toPorts:
+        - ports:
+            - port: "8000"
+              protocol: TCP
```
pwoelfle-unique: Can be removed since kong ingress rule is done via `internalServices.dependents.ingressGateway` and ingress from host is not necessary.

```suggestion
```

nicolae536 (approved):

services/hello-mcp/deploy/helm-charts/hello-mcp/values.yaml
proschu2: Removed — Kong via `internalServices.dependents.ingressGateway`, no manual host/Kong `networkPolicy.ingress` rules.

services/hello-mcp/deploy/helm-charts/hello-mcp/values.yaml
```diff
@@ -0,0 +1,105 @@
...
+  enabled: false
+  flavor: cilium
+  enableDefaultDeny:
+    ingress: true
+    egress: true
+  baseline:
+    prometheus:
+      enabled: false
+    ephemeralPorts:
+      enabled: false
```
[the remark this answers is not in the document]
proschu2: Fixed — added `volumes`/`volumeMounts` emptyDir at `/tmp` (64Mi) to match sibling MCP charts under `readOnlyRootFilesystem`.

konsti (approved):
I just reviewed the file structure and pnpm compat setup. :shipit:

pwoelfle-unique (approved):

PR #716, opened 2026-07-30
feat(backstop-mcp,ci,scripts,main): scaffold Backstop MCP service shell

nicolae536:
Scaffold the Backstop MCP service: packaging, Postgres/alembic, Helm, and CI. Health/probe/ready only — no tools yet.

Refs: UN-22647


release-please-config.json
```diff
@@ -121,6 +121,25 @@
           "path": "deploy/helm-charts/hello-mcp/values.yaml"
         }
       ]
+    },
+    "services/backstop-mcp": {
```
proschu2, line 126: add component, see updated config

docs/plans/2026-07-27-backstop-mcp-scaffold-design.md
```diff
@@ -0,0 +1,61 @@
+# Design: backstop-mcp scaffold
```
proschu2: not sure u wanted to push plans?

release-please-config.json
nicolae536: Added `"component": "backstop-mcp"` to match the updated hello-mcp release-please config.

docs/plans/2026-07-27-backstop-mcp-scaffold-design.md
nicolae536: Agreed — removed `docs/plans/` and `docs/pr-proposals/` from the PR; those were local planning artifacts.

services/backstop-mcp/src/backstop_mcp/auth/provider.py
```diff
@@ -0,0 +1,448 @@
...
+                    .where(
+                        OAuthTokenRow.family_id == row.family_id,
+                        OAuthTokenRow.revoked_at.is_(None),
+                    )
+                    .values(revoked_at=now)
+                )
+            else:
+                new_access_token = secrets.token_urlsafe(32)
+                new_refresh_token = secrets.token_urlsafe(32)
+                effective_scopes = scopes or row.scopes
```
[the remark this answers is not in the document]
nicolae536: Fixed — refresh now rejects requested scopes that are not a subset of the originally granted scopes (`invalid_scope`), and the original refresh token stays usable after a rejected escalation.

services/backstop-mcp/src/backstop_mcp/db/models.py
```diff
@@ -0,0 +1,118 @@
...
+    created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now())
+
+
+class BackstopCredential(Base):
+    """A user's Backstop username + personal API token, encrypted at rest."""
+
+    __tablename__: str = "backstop_credentials"
+
+    user_id: Mapped[str] = mapped_column(String, primary_key=True)
+    backstop_username: Mapped[str] = mapped_column(String, index=True)
```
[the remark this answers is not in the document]
nicolae536: Fixed — `backstop_username` is now unique (model + initial migration) so concurrent/reconnect identity lookup stays one row per username.

services/backstop-mcp/src/backstop_mcp/auth/provider.py
```diff
@@ -0,0 +1,512 @@
...
+                    subject=user_id,
+                    expires_at=code_expires_at,
+                )
+            )
+            await session.execute(
+                delete(PendingAuthorization).where(PendingAuthorization.request_id == request_id)
+            )
+
+        redirect_url = construct_redirect_uri(pending.redirect_uri, code=code, state=pending.state)
+        return RedirectResponse(redirect_url, status_code=302)
```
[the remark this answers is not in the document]
nicolae536: Fixed — `handle_login_post` now claims the pending authorization with an atomic `DELETE ... WHERE request_id = ...` before minting the auth code (same pattern as code exchange / refresh rotation). Concurrent submits for the same request only succeed once; the loser gets the expired-link response. Covered by `test_concurrent_login_only_mints_one_code`.

services/backstop-mcp/tests/test_backstop_client.py
```diff
@@ -0,0 +1,885 @@
...
+            return_value=httpx.Response(
+                200,
+                json={"data": [{"id": "1"}], "links": {"next": "https://evil.example.com/records"}},
+            )
+        )
+
+        with pytest.raises(BackstopUntrustedUrlError) as exc_info:
+            await factory.for_credential(_credential()).paginate("/records")
+
+        assert "evil.example.com" in str(exc_info.value)
```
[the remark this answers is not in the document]
nicolae536: Fixed — the assertion now checks `BackstopUntrustedUrlError.url` / `.expected_host` instead of a hostname substring in the message, which clears the Incomplete URL substring sanitization CodeQL alert.

services/backstop-mcp/src/backstop_mcp/backstop_client/json_api.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    wanted = resource.related_ids(relationship_name)
+    if not wanted:
+        return []
+    by_id = {
+        item_id: item
+        for item in document.included
+        if (item_id := as_clean_str(item.get("id"))) is not None
+    }
+    return [by_id[related_id] for related_id in wanted if related_id in by_id]
```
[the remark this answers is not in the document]
nicolae536: Fixed — `included_for_relationship` now matches `included` entries by JSON:API identity `(type, id)` instead of id alone, so colliding numeric ids across `entity-relationships` / `entity-relationship-types` no longer overwrite each other. Covered by `test_matches_by_type_and_id_when_ids_collide_across_types`.

docs/plans/2026-08-05-employment-index-design.md
```diff
@@ -0,0 +1,202 @@
...
+4. **Add the two builders** — `build_person_employment_index` and
+   `build_organization_employment_index`, differing only in which side holds the person.
+5. **Rework the service** — `EmploymentIndexFactory` with `index_for_person` and
+   `index_for_organization`; rename the factory function and the runtime accessor and update
+   `create_app()`.
+6. **Update `get_person`** — build the person index and replace `departed_detail` with
+   `departures: list[DepartedContactEcho]`, keeping `departed: bool` as "any departure."
+7. **Document the mirror types in `.env.example`** — id-based `FORMER` configuration needs the
+   `(mirror)` type's id as well; the name markers already cover both.
+8. **Write the behavioural tests** — the nine cases above.
```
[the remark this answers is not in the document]
nicolae536: Removed again — deleted `docs/plans/` and `docs/pr-proposals/` (reintroduced with the employment-index work) and dropped the test docstring reference to the plan.

docs/plans/2026-08-03-backstop-client-schema-deserialization-design.md
```diff
@@ -0,0 +1,198 @@
...
+
+6. **Define `PartyAttributes` and migrate `party_resolver/search.py`** - replace
+   `_candidate_from_resource`/`_display_name`/`_as_object_dict` with
+   `GetRequest(schema=JsonApiDocument[PartyAttributes])`.
+7. **Define `OrganizationAttributes` and migrate `tools/get_organization.py`** - replace
+   `_name_from_organization_body` with typed attribute access; rebuild the `organization`
+   dict via `model_dump(exclude_none=True)`.
+8. **Update existing party_resolver / get_organization tests** - cover the new
+   raise-on-malformed-item behavior; confirm the org round-trip assertion
+   (`result["organization"] == org_body`) still passes byte-for-byte.
```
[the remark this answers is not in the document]
nicolae536: Removed — `docs/plans/` and `docs/pr-proposals/` are out of the PR again.

docs/plans/2026-07-27-backstop-mcp-scaffold-design.md
```diff
@@ -0,0 +1,61 @@
...
+
+No integration/testcontainers setup — there's no DB or queue to spin up.
+
+## Out of Scope
+
+- Any Backstop domain tools (contacts, opportunities, accounts, activities, targeting).
+- Auth against the Backstop REST API.
+- Party-ID data model / any Backstop-specific schema.
+- RabbitMQ or any event/write-back queue plumbing.
+- Postgres or any caching/staleness-tracking layer.
```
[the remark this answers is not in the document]
nicolae536: Those early scaffold plan docs are no longer in the tree (planning artifacts removed). The implementation (Postgres/OAuth/credentials) is the source of truth.

services/backstop-mcp/src/backstop_mcp/features/auth/provider.py
```diff
@@ -0,0 +1,448 @@
...
+        """
+        async with get_session(self._session_factory) as session:
+            await session.execute(
+                update(OAuthTokenRow)
+                .where(
+                    OAuthTokenRow.subject == subject,
+                    OAuthTokenRow.revoked_at.is_(None),
+                )
+                .values(revoked_at=datetime.now(UTC))
+            )
```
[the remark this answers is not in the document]
nicolae536: Already fixed — Backstop 401 calls `on_auth_failure` → `revoke_current_subject_tokens` → `revoke_token_family_for_subject` (wired in `factory.for_current_caller` / `app.py`). Covered by client + system_info auth tests.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,142 @@
...
+        return values
+
+    @model_validator(mode="after")
+    def build_url(self) -> "DatabaseConfig":
+        if self.url is not None:
+            if not self.url.startswith(("postgresql://", "postgresql+")):
+                raise ValueError("DB_URL must be a PostgreSQL connection string (postgresql://...)")
+            if self.url.startswith("postgresql://"):
+                self.url = self.url.replace("postgresql://", "postgresql+asyncpg://", 1)
+            return self
```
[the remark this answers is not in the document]
nicolae536: Already fixed — `normalize_asyncpg_url` strips libpq `sslmode`/`channel_binding` and maps SSL into asyncpg `connect_args`. Covered by `tests/test_config.py`.

services/backstop-mcp/src/backstop_mcp/features/auth/provider.py
```diff
@@ -0,0 +1,458 @@
...
+                error_description="Requested scope exceeds originally granted scopes",
+            )
+
+        return OAuthTokenResponse(
+            access_token=new_access_token,
+            token_type="Bearer",
+            expires_in=int(self.ACCESS_TOKEN_TTL.total_seconds()),
+            scope=" ".join(effective_scopes) if effective_scopes else None,
+            refresh_token=new_refresh_token,
+        )
```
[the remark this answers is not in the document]
nicolae536: Already fixed — `load_refresh_token` intentionally returns rotated (revoked) tokens so reuse detection in `_rotate_refresh_token` can revoke the family. Covered by `test_refresh_token_rotates_and_detects_reuse`.

services/backstop-mcp/src/backstop_mcp/auth/context.py
```diff
@@ -0,0 +1,77 @@
...
+from backstop_mcp.auth.crypto import BackstopCredentialSecret
+from backstop_mcp.db.engine import get_session
+
+
+class NotConnectedError(ToolError):
+    """Raised when the caller isn't authenticated, or has no Backstop credential on file.
+
+    Surfaced to the MCP client as a tool error telling them to (re)connect this server —
+    see `docs/plans/2026-07-27-backstop-mcp-scaffold-design.md` for the login flow this refers to.
+    """
```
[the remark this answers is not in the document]
nicolae536: Already fixed — `NotConnectedError` no longer references `docs/plans/…`.

services/backstop-mcp/src/backstop_mcp/features/auth/provider.py
```diff
@@ -0,0 +1,458 @@
...
+                    access_token_expires_at=now + self.ACCESS_TOKEN_TTL,
+                    refresh_token_expires_at=now + self.REFRESH_TOKEN_TTL,
+                )
+            )
+            # Authorization codes are single-use.
+            await session.execute(
+                delete(AuthorizationCodeRow).where(
+                    AuthorizationCodeRow.code == authorization_code.code
+                )
+            )
```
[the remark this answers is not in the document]
nicolae536: Already fixed — `exchange_authorization_code` atomically claims the code with `DELETE … WHERE code = …` before minting tokens. Covered by `test_concurrent_authorization_code_exchange_only_succeeds_once`.

services/backstop-mcp/src/backstop_mcp/auth/provider.py
```diff
@@ -0,0 +1,458 @@
...
+                            refresh_token_hash=_hash_token(new_refresh_token),
+                            client_id=row.client_id,
+                            scopes=effective_scopes,
+                            resource=row.resource,
+                            subject=row.subject,
+                            access_token_expires_at=now + self.ACCESS_TOKEN_TTL,
+                            refresh_token_expires_at=now + self.REFRESH_TOKEN_TTL,
+                            rotated_from=row.id,
+                        )
+                    )
```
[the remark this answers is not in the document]
nicolae536: Already fixed — refresh rotation claims with `UPDATE … WHERE id = ? AND revoked_at IS NULL`; concurrent losers hit the reuse/revoke-family path. Covered by `test_concurrent_refresh_of_same_token_never_leaves_two_valid_descendants`.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,326 @@
...
+                # for those, which the retry predicate in retry.py inspects.
+                raise parse_json_api_error(response)
+            return response
+
+        logger.debug("backstop.request.start", method=method, path=path)
+        try:
+            response: httpx.Response = await retrying(make_request)
+        except Exception as exc:
+            logger.error("backstop.request.failed", method=method, path=path, error=str(exc))
+            raise
```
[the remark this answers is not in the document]
nicolae536: Already fixed — request failures use `logger.exception` / structured `BackstopApiError` fields; unparseable bodies log `body_length` rather than raw `response.text`.

services/backstop-mcp/src/backstop_mcp/features/auth/provider.py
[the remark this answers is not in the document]
nicolae536: Already fixed — expired refresh tokens are rejected in `_rotate_refresh_token`, and auth-code/refresh expiry is surfaced to the MCP TokenHandler via `expires_at`. Covered by `test_refresh_token_rejects_expired_token`.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+from backstop_mcp.backstop_client.settings import BackstopTransportSettings
+from backstop_mcp.metrics import (
+    BACKSTOP_CONCURRENCY_WAIT,
+    BACKSTOP_REQUEST_DURATION,
+    BACKSTOP_REQUESTS,
+)
+
+logger = logging.getLogger(__name__)
+
+_AUTHORIZATION_HEADER = "authorization"
```
klcd: Inline?

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+# collaborators. It resolves the caller's stored credential, so calling it costs a DB read plus
+# a decrypt: acquire one only on a path that will actually make a request.
+type CallerClientProvider = Callable[[], Awaitable["BackstopClient"]]
+# Acquired around a *single* upstream request (see `BackstopClient._request`), never around a
+# whole tool invocation — Backstop's limit is on concurrent requests, and a caller that holds
+# a slot across an elicitation prompt or a batch of gathered calls either starves itself or
+# breaches the limit. See `BackstopClientFactory`.
+type RequestGate = Callable[[str], AbstractAsyncContextManager[None]]
+
+_DICT_ADAPTER = TypeAdapter(dict[str, object])
```
klcd: This exists because this is not a class

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+        raise BackstopResponseSchemaError(path, schema.__name__, exc) from exc
+
+
+# /reports and /{entity}/{id}/analytics are the calls Backstop docs call out as legitimately
+# slow (up to ~30s per 500 records) — they get the extended timeout and the larger
+# report-sized page default; everything else gets the ordinary CRUD profile.
+_EXTENDED_PROFILE_MARKERS = ("/reports", "/analytics")
+
+
+class BackstopAuthError(Exception):
```
klcd: Separate exceptions file

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+    """Build the `Authorization: Basic ...` + `token: true` headers Backstop expects.
+
+    Every user connects with a personal API token (not a password), so `token: true` is
+    always sent — see https://backstopsolutions.elevio.help/en/articles/1018 and .../236.
+    """
+    basic_auth = base64.b64encode(f"{username}:{api_token}".encode()).decode()
+    return {_AUTHORIZATION_HEADER: f"Basic {basic_auth}", _TOKEN_HEADER: "true"}
+
+
+def is_extended_profile_path(path: str) -> bool:
```
klcd: I would say any function that is a oneliner and not used 5 times is unecessary

klcd: Just used locally, so make it a private?

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+    return any(marker in path for marker in _EXTENDED_PROFILE_MARKERS)
+
+
+def build_url(base_url: str, path: str) -> str:
```
klcd: from pathlib import Path helps with actual file paths

klcd: also there is urlib

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,209 @@
...
+                "DB_USER": self.user,
+                "DB_PASSWORD": self.password,
+            },
+            url_field="DB_URL",
+        )
+
+        assert self.user is not None
+        assert self.password is not None
+        self.url = (
+            f"postgresql+asyncpg://{quote_plus(self.user)}:{quote_plus(self.password)}"
```
klcd: Assembling a DSN by f-string is the thing SQLAlchemy's `URL` exists to prevent, and it's already a dependency:

```python
from sqlalchemy import URL

self.url = URL.create(
    "postgresql+asyncpg",
    username=self.user,
    password=self.password,
    host=self.host,
    port=self.port,
    database=self.name,
).render_as_string(hide_password=False)
```

Two concrete things this fixes:

- `quote_plus` is the wrong escaper. It's percent-encoding for `application/x-www-form-urlencoded`, so it turns a space into `+`, not `%20` — a password containing a space silently becomes the wrong password. `URL.create` takes the raw value and escapes per RFC 3986.
- `self.host` and `self.name` are interpolated with no escaping at all. An IPv6 host (`::1`) produces a malformed DSN, and a database name with a `/` or `?` in it changes what's being parsed.

As a bonus, keeping it as a `URL` object rather than re-rendering to a string means `db/engine.py` can pass it straight to `create_async_engine`.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,209 @@
...
+    if url.startswith("postgresql://"):
+        url = url.replace("postgresql://", "postgresql+asyncpg://", 1)
+
+    parsed = urlparse(url)
+    params = dict(parse_qsl(parsed.query, keep_blank_values=True))
+    sslmode = params.pop("sslmode", None)
+    # Also libpq-only; asyncpg does not accept it in the query string.
+    params.pop("channel_binding", None)
+
+    clean_url = urlunparse(parsed._replace(query=urlencode(params)))
```
klcd: This whole function is URL surgery with `urlparse` / `parse_qsl` / `namedtuple._replace` / `urlencode` — to change a driver name and drop two query params. SQLAlchemy has both operations as first-class methods:

```python
from sqlalchemy import make_url

url = make_url(raw)
sslmode = url.query.get("sslmode")
url = url.set(drivername="postgresql+asyncpg").difference_update_query(
    ["sslmode", "channel_binding"]
)
```

That also removes the `startswith("postgresql://")` / `.replace(..., 1)` prefix juggling on line 59 (`.set(drivername=...)` doesn't care what it was) and the separate `startswith(("postgresql://", "postgresql+"))` guard on line 175 — `make_url` raises on anything it can't parse, and you can check `url.get_backend_name() == "postgresql"` for the rest.

Worth noting `parsed._replace` is a namedtuple method with a leading underscore. It's public API despite the spelling, but it reads like reaching into internals, which is one more reason to let SQLAlchemy own this.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,209 @@
...
+            return data
+        values = cast(dict[str, object], data)
+        if values.get("url") is None:
+            database_url = os.environ.get("DATABASE_URL")
+            if database_url:
+                return {**values, "url": database_url}
+        return values
+
+    @model_validator(mode="after")
+    def build_url(self) -> "DatabaseConfig":
```
klcd: Naming: `build_url` is a validator, not a builder you call — and there's an unrelated `build_url` over in `backstop_client/client.py` that resolves HTTP paths. Two functions, one name, different domains.

Given it also handles the parse-the-DSN branch, `_resolve_connection_url` (leading underscore, like `_require_complete_service_account` below) says what it is and stops it reading as public API.

On the shape: a `mode="after"` validator that reassigns `self.url` and `self._connect_args` is pydantic-legal but means the model has two lifecycle states, which is what forces the `RuntimeError` guard on line 203. A `@computed_field` (or a plain cached property) derived from the discrete inputs would have one.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,209 @@
...
+            f"@{self.host}:{self.port}/{self.name}"
+        )
+        self._connect_args = AsyncpgConnectArgs()
+        return self
+
+    @property
+    def connection_url(self) -> str:
+        """Get the connection URL (guaranteed non-None after validation)."""
+        if self.url is None:
+            raise RuntimeError("URL should be set after validation")
```
klcd: `url: str | None` plus a property that raises `RuntimeError("URL should be set after validation")` is the model telling you the type doesn't match reality — after validation it is never `None`, but nothing expresses that, so every reader gets an `Optional` they have to defend against.

Two options, both better than the guard: keep `url` private (`_url: str = PrivateAttr()`) and let `connection_url` return `str` outright, or model the two input shapes as a union (`DatabaseUrlConfig | DatabasePartsConfig`) and let pydantic pick.

Smaller point, per `CLAUDE.md`: internal invariants like this use `assert` from `node:assert`'s Python equivalent — a bare `assert self.url is not None, "..."` — rather than `if` + `raise`. Lines 190-191 already do exactly that a few lines up.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,209 @@
...
+_SSLMODE_VERIFY = frozenset({"verify", "verify-ca", "verify-full"})
+# Encrypt the connection but do not verify the server certificate.
+_SSLMODE_ENCRYPT = frozenset({"require", "prefer"})
+
+
+class AsyncpgConnectArgs(TypedDict, total=False):
+    ssl: ssl.SSLContext
+
+
+def _validate_required_fields(fields: dict[str, str | None], url_field: str = "URL") -> None:
```
klcd: This reimplements what pydantic does natively — collect the missing required fields and raise with all of them listed. And the `url_field` parameter isn't a field name, it's a prefix for the error message, which is why its default is the string `"URL"` and its only caller passes `"DB_URL"`.

If the four fields are genuinely required in the no-DSN case, model that case as its own `BaseModel` and let pydantic report the missing ones with proper locations and error codes. Failing that, inline the check into `_resolve_connection_url` — it's four lines and one caller.

While here: `fields`, `val`, `missing` are fine, but `url_field` should be `error_label` if it survives.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,209 @@
...
+
+    Credentials are NOT configured here: each connecting MCP client will eventually complete a
+    hosted login form, verified against Backstop and stored encrypted in Postgres. That bridge —
+    and the tuning knobs for the shared HTTP client — land in a later PR; for now this only
+    carries the base URL tools will eventually talk to.
+    """
+
+    model_config: ClassVar[SettingsConfigDict] = SettingsConfigDict(env_prefix="BACKSTOP_")
+
+    base_url: str = "https://api.backstopsolutions.com"
```
klcd: `base_url: str` on a pydantic-settings model means an operator can set `BACKSTOP_BASE_URL=not a url` and the failure surfaces later as a confusing httpx error rather than at startup, where every other bad value in this file is caught.

`HttpUrl` validates and normalizes at load. That matters downstream: three separate places in #748 do `base_url.rstrip('/')` because a string may or may not have a trailing slash, and a normalized URL removes the need for any of them.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,209 @@
...
+        """Fail fast when a production deploy never set `PUBLIC_BASE_URL`.
+
+        Left at the default, this service advertises a loopback issuer in its OAuth discovery
+        metadata and redirects browsers to a login form on the client's own machine. Nothing
+        errors server-side — clients just fail to connect for a reason nothing here reports. So
+        it's rejected at startup, the same way a missing encryption key is.
+        """
+        if self.app_env != AppEnv.PRODUCTION:
+            return self
+        host = urlparse(self.public_base_url).hostname
```
klcd: Same as `base_url`: type this as `HttpUrl` and the `urlparse(...)` here goes away — you'd read `self.public_base_url.host` directly, and the `host is None` branch disappears because a validated `HttpUrl` always has one.

It also removes the third parse of this same value: `provider.py` in #748 does `urlparse(base_url).scheme == "https"` to decide the cookie `Secure` flag, and `authorize()` does `str(self.base_url).rstrip("/")`. One validated URL object serves all three.

services/backstop-mcp/src/backstop_mcp/coerce.py
```diff
@@ -0,0 +1,31 @@
...
+    return cast("dict[str, object]", value)
+
+
+def as_object_list(value: object) -> list[object]:
+    if not isinstance(value, list):
+        return []
+    return cast("list[object]", value)
+
+
+def as_clean_str(value: object) -> str | None:
```
klcd: `as_clean_str` — strip, then `None` if empty — is `Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]`, which the codebase already uses over in `backstop_client/json_api.py`. Same rule written two ways, depending only on whether the value arrived through a pydantic model or through a `dict[str, object]`.

The deeper point is that `as_object_dict` / `as_object_list` exist because those payloads were never modelled. The docstring says they're for "corners of Backstop's API that the swagger declares as free-form objects" — but `lovSet` and `selectOptions` do have a runtime shape, and one `TypeAdapter` for it would replace all three helpers plus the `cast()`s they wrap.

Naming: the module is `coerce` but nothing here coerces — these narrow, returning `None`/`[]` on a miss. `narrow.py`, or fold them into models and delete the module.

klcd (commented):
Follow-up on the Python in the scaffolding, same lens as my comments on #748: things that are hand-rolled where a dependency already in `pyproject.toml` does it, plus naming.

`config.py` is the concentration. It builds and rewrites Postgres DSNs with `urllib.parse` string surgery and f-strings while SQLAlchemy — which is the thing consuming those DSNs — exposes `URL` for exactly this, and it types every URL as a bare `str` while pydantic ships `HttpUrl`/`PostgresDsn`.

(My earlier comments on this PR were left before the stack was split; the ones about `backstop_client/client.py` now live on #748.)

services/backstop-mcp/tests/test_app.py
```diff
@@ -0,0 +1,292 @@
...
+            lambda: AccessToken(
+                token="access-token",
+                client_id="client-1",
+                scopes=[],
+                subject="user-never-connected",
+            ),
+        )
+
+        with pytest.raises(NotConnectedError):
+            await get_services().backstop.for_current_caller()
```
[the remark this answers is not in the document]
nicolae536: Correcting myself — you were right, and my earlier reply here was wrong.

I checked this by running the test in isolation, where it passes in ~3s. That cannot rule out an order-dependent race, and in a full-suite run it doesn't just flake: it **hangs indefinitely**. The suite stopped dead at 23%, and a `faulthandler` dump showed the main thread inside pytest-asyncio with its event loop parked in `selectors.select()` — awaiting a future nothing will complete — while the `TestClient` blocking-portal thread was still alive.

Mechanism is exactly as you described: `TestClient` runs the lifespan on a portal thread and loop, `cleanup_lifespan` sweeps on start and so binds the engine's asyncpg pool to that loop, and the test body then awaits a credential lookup from pytest-asyncio's function-scoped loop.

Fixed in 800e0d8b — the test builds the app with `create_app` instead of taking `app_client`. It only ever needed the cycle to have been closed, not a running server, so nothing touches the engine before its own loop does. Full suite now completes in ~6s.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Gone — the `authorization` / `token` header constants this anchored on no longer exist: auth is per-request `httpx.BasicAuth`, and the static headers live once in `factory._SHARED_CLIENT_HEADERS`. `client.py` now keeps only `_errors_for_log` at module level. Landed in #748 (69071378).

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Agreed — `client.py` is now a `BackstopClient` class, and the module-level `_DICT_ADAPTER` / `_deserialize` pair this anchored on is gone (`schema` is required, so there's no untyped branch left to need it). The remaining pure helpers moved to `backstop_client/utils.py`. Landed in #748 (69071378).

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Done — `BackstopAuthError` and `BackstopUnreachableError` live in `backstop_client/errors.py`; `client.py` defines no exception classes. Landed in #748 (69071378), where you re-anchored this.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Folded in, as offered — fixed in 8edc0281. `is_slow_endpoint` is gone; both call sites inline `any(marker in path for marker in _SLOW_ENDPOINT_MARKERS)`, and the marker tuple sits next to them in `client.py` with the comment that explains it.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Took your re-anchored correction on #748 — `pathlib` would mangle `https://` into `https:/`, so it's `httpx` instead. The hand-rolled `build_url` is gone: the factory builds `httpx.AsyncClient(base_url=...)` so relative paths merge in httpx, and `utils.resolve_request_url` uses `httpx.URL` to pin an absolute `links.next` to the configured scheme *and* netloc. No `pathlib` or `urllib` left in the URL path. Landed in #748 (69071378 + 453f646d).

services/backstop-mcp/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Migrations
+
+```bash
+uv run alembic upgrade head                          # apply
+uv run alembic revision --autogenerate -m "message"  # create
+uv run alembic downgrade -1                          # roll back one
+```
+
+There are no versions under `db/migrations/versions/` yet — `upgrade head` is a no-op until the
+first feature adds tables.
```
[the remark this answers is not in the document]
nicolae536: Fixed — `src/backstop_mcp/db/migrations/versions/` is tracked (a `.gitkeep` plus four real migrations), so `alembic upgrade head` works from a fresh clone. The README no longer claims it's a no-op either: it lists `uv run alembic upgrade head` as a setup step, and `tests/conftest.py` runs the migrations against the Postgres container.

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: Done in 856aed37 — the DSN is `URL.create("postgresql+asyncpg", username=..., password=..., host=..., port=..., database=...).render_as_string(hide_password=False)`. `quote_plus` is gone from the service entirely, so the space-becomes-`+` password bug goes with it.

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: Done in 856aed37 — `normalize_asyncpg_url` is now `make_url(url).set(drivername="postgresql+asyncpg").difference_update_query(["sslmode", "channel_binding"])`. No `urlparse` / `parse_qsl` / `_replace` / `urlencode` left, and the `startswith("postgresql://")` prefix juggling is gone; what remains is a drivername guard on the already-parsed URL.

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: Done in 856aed37 — renamed to `_resolve_connection_url`. The collision is gone on both sides: `build_url` no longer exists anywhere in the service, since the client-side URL joining became `utils.resolve_request_url`.

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: Done in 856aed37 — took the `PrivateAttr` option: `_connection_url: str = PrivateAttr(default="")`, and `connection_url` returns `str` outright. The `RuntimeError` guard is gone (no `RuntimeError` left in `config.py` at all). `url: PostgresDsn | None` stays, but that's the genuinely-optional raw env input rather than the accessor you were objecting to.

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: Done in 856aed37 — took the inline branch: the helper and its `url_field` parameter are gone, and the missing-field collection is now four lines inside `_resolve_connection_url`. That does mean pydantic still isn't the one reporting them with proper locations/codes — happy to go the "model the no-DSN case as its own BaseModel" route if you'd prefer it.

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: The leftover is gone too, in 6f1f9525 — `CustomFieldsService` no longer re-trims `base_url`. It is normalised once by `BackstopConfig.base_url` and is half of the snapshot key, so re-trimming there would only have hidden a base URL that disagreed with it.

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: Now fixed in 04c3241c / 8edc0281 — all three parses are gone.

`public_base_url` is typed `HttpUrl`, so the production check reads `.host` off it directly and the `host is None` branch is an assert on an invariant rather than a live path. `provider.py` no longer imports `urlparse`: the cookie `Secure` flag is passed in as `secure_cookies` from the one place the URL is parsed, and `authorize()` uses the already-normalised issuer string instead of `str(self.base_url).rstrip("/")`.

`AppConfig.issuer` is where the trailing slash comes off, once. Worth noting for the record: the discovery document is not that value — the MCP SDK re-parses the issuer into an `AnyHttpUrl` and advertises it with the slash restored, and it strips it again itself to build the authorize/token URLs.

services/backstop-mcp/src/backstop_mcp/coerce.py
nicolae536: Now done properly, across the stack — `coerce.py` is deleted.

- `as_clean_str`: the rule moved onto the fields that produce the values. `data_hygiene/types.py` has a single `CleanStr` annotation, so `provenance.py` and `employment.py` read already-clean values instead of re-checking at each use.
- `as_object_dict` / `as_object_list`: `lovSet` / `selectOptions` are modelled now. `_InlineOption` normalises the bare-string, flat-object and JSON:API-resource spellings; `_InlineOptionSet` resolves the collection-key and `data` shapes. `inline_allowed_values` is a concatenation rather than a shape probe.
- The module is gone, so the naming point is moot and `lov.py` no longer redeclares its own `_CleanStr`.

One trap worth recording, since it's the same one you caught on #749: the `CleanStr` annotation has to sit on the union, not on the `str` arm — inside `Annotated[str, ...] | None` a blank converted to `None` is still checked against `str` and rejected.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,243 @@
...
+)
+from pydantic_settings import BaseSettings, SettingsConfigDict
+from sqlalchemy.engine.url import URL, make_url
+
+PKG_VERSION = pkg_version("backstop-mcp")
+
+# libpq sslmode values that require certificate verification.
+_SSLMODE_VERIFY = frozenset({"verify", "verify-ca", "verify-full"})
+# Encrypt the connection but do not verify the server certificate.
+_SSLMODE_ENCRYPT = frozenset({"require", "prefer"})
```
klcd: Only used once

klcd: both of them

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,243 @@
...
+
+    base_url: HttpUrlStr = "https://api.backstopsolutions.com"
+
+
+class DatabaseConfig(BaseSettings):
+    """Where backstop-mcp stores its state."""
+
+    model_config: ClassVar[SettingsConfigDict] = SettingsConfigDict(env_prefix="DB_")
+
+    url: PostgresDsn | None = None
```
klcd, line 189: Can these values be `None` and the thing works?

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: Inlined both `_SSLMODE_*` frozensets into the `in (...)` checks — each was only used once.

services/backstop-mcp/src/backstop_mcp/config.py
nicolae536: Yes when `DB_URL` / `DATABASE_URL` is set — the discrete fields are only required when building a DSN from parts. Clarified that dual-mode contract on `DatabaseConfig`.

services/backstop-mcp/src/backstop_mcp/config.py
```diff
@@ -0,0 +1,244 @@
...
+    if sslmode in ("disable", "allow"):
+        return None
+    if sslmode in ("require", "prefer"):
+        # Encrypt the connection but do not verify the server certificate.
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+        return ctx
+    if sslmode in ("verify", "verify-ca", "verify-full"):
+        return ssl.create_default_context()
```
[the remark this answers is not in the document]
nicolae536: Fixed in ca54e93a — `sslmode=verify-ca` now builds a context with `CERT_REQUIRED` and `check_hostname=False`; `verify` / `verify-full` keep the default hostname check. Covered by `test_strips_sslmode_verify_ca_without_hostname_check`.

vabatta (approved):

services/backstop-mcp/pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+    "python-dotenv==1.2.2",
+    "httpx==0.28.1",
+    "opentelemetry-api==1.44.0",
+    "opentelemetry-sdk==1.44.0",
+    "opentelemetry-exporter-prometheus==0.65b0",
+    "opentelemetry-instrumentation-asgi==0.65b0",
+    "sqlalchemy[asyncio]==2.0.51",
+    "asyncpg==0.31.0",
+    "alembic==1.18.5",
+    "cryptography==49.0.0",
```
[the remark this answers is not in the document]
nicolae536: Fixed: pinned `cryptography==50.0.0` (same as kb-mcp) and regenerated `uv.lock`.

PR #719, opened 2026-07-30
feat(kb-mcp,ci,scripts,main): migrate Knowledge Base Search MCP into connectors as kb-mcp

proschu2:
## Summary
- Migrate Unique KB MCP (`mcp_search`) into connectors as deployable FastMCP service **`kb-mcp`** (directory, chart, image, Python package `kb_mcp`, CLI; product wording remains “Knowledge Base Search”)
- Zitadel OIDC via `unique-mcp`; KB tools (`search`, `content_tree`, `read_file`) via `unique-toolkit`
- Ops: pino-json logging (`configure_logging` + Loki pod label), `/probe` + Prometheus (`setup_ops`), opt-in OTEL tracing (`configure_tracing`)
- Pins: `unique-mcp==2026.34.0.dev3`, `unique-toolkit[monitoring,otel]==2026.34.0.dev5`, `unique-sdk==2026.34.0.dev2`, `fastmcp==3.4.5`
- Helm chart + Dockerfile + release-please / CD dispatch under `services/kb-mcp` (Gateway API routes, secrets via overlay `envVars`)

## Follow-ups (deploy / monorepo)
- Argo overlay still needs `UNIQUE_MCP_PUBLIC_BASE_URL`, Gateway `routes.hostname`, and secret `envVars` (not baked into chart defaults)
- OTLP remains opt-in (`OTEL_*`); wire collector endpoint in the overlay when ready

## Test plan
- [x] `uv sync` + `pytest` in `services/kb-mcp` (65 passed)
- [x] `helm template` for `services/kb-mcp/deploy/helm-charts/kb-mcp`
- [x] Python CI matrix discovers `kb-mcp` via `pyproject.toml`
- [ ] Manual smoke: OAuth + `/mcp` against a tenant


services/kb-mcp/src/kb_mcp/main.py
```diff
@@ -0,0 +1,81 @@
...
+        instructions=SERVER_CITATION_INSTRUCTIONS,
+        auth=oidc_proxy,
+        providers=[FileSystemProvider(Path(__file__).parent / "tools")],
+    )
+
+    middleware = [
+        Middleware(
+            CORSMiddleware,
+            allow_credentials=True,
+            allow_origins=["*"],
```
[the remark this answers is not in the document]
proschu2: Fixed in 466de79 — dropped `allow_credentials=True`. MCP auth is Bearer/`_meta` in the request body, not cookies; wildcard + credentials made Starlette reflect any `Origin`. OAuth redirects are top-level and unaffected.

services/kb-mcp/src/kb_mcp/scope_resolver.py
```diff
@@ -0,0 +1,78 @@
...
+        if scope:
+            resolved[chunk.id] = scope
+        else:
+            missing.add(chunk.id)
+
+    if not missing:
+        return resolved
+
+    user_id = settings.auth.get_confidential_user_id()
+    company_id = settings.auth.get_confidential_company_id()
```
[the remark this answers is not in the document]
proschu2: Fixed in 466de79 — `resolve_scope_ids` now uses `settings.authcontext` like `content_tree` / `read_file`. (`settings.auth` still works but is deprecated.)

klcd (approved):

klcd (approved):

services/kb-mcp/src/kb_mcp/tools/read_file.py
```diff
@@ -0,0 +1,310 @@
...
+    if result.isError or not result.content:
+        return result
+    first = result.content[0]
+    if not isinstance(first, TextContent):
+        return result
+    url = file_reference_url(
+        content.id,
+        metadata=content.metadata,
+        frontend_base_url=KbMcpServerSettings().frontend_base_url_str(),
+    )
```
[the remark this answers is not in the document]
proschu2: Dismissing: toolkit `Content` has no top-level `owner_id` (`map_content` drops API `ownerId`); only `ContentInfo` has it. `read_file` already passes `metadata=content.metadata` into `file_reference_url`, and `scope_id_from_metadata` already handles `folderIdPath` / metadata `ownerId` — enough for normal KB docs. Fetching `ContentInfo` via `get_info_async` for orphan deep-link parity is optional and not worth an extra call; `unique://` fallback is fine.

services/kb-mcp/src/kb_mcp/main.py
```diff
@@ -0,0 +1,82 @@
...
+    # own names — loading into the process env covers all of them).
+    # override=True so service .env wins over stale shell UNIQUE_MCP_* exports.
+    load_dotenv(Path.cwd() / ".env", override=True)
+    # Opt-in via OTEL_* (e.g. OTEL_TRACES_EXPORTER=console locally).
+    configure_tracing(service_name="kb-mcp")
+    configure_logging()
+
+    server_settings = ServerSettings()
+
+    oidc_proxy = create_zitadel_oidc_proxy(
```
konsti, line 32: I'm not a fan of this Zitadel OIDC proxy. The regular FastMCP proxy works totally fine. It obscures authentication behind code in a different repo. Hard for agents/humans to review.

services/kb-mcp/src/kb_mcp/main.py
```diff
@@ -0,0 +1,82 @@
...
+from unique_toolkit.monitoring import configure_tracing
+
+from kb_mcp.references import SERVER_CITATION_INSTRUCTIONS
+
+
+def main() -> None:
+    # Single .env for local (toolkit prefers unique.env; mcp/zitadel prefer their
+    # own names — loading into the process env covers all of them).
+    # override=True so service .env wins over stale shell UNIQUE_MCP_* exports.
+    load_dotenv(Path.cwd() / ".env", override=True)
```
konsti, line 25: Rather than this I would have a central config module and load / validate config there. In production environments such code must be disallowed.

See in my branch how I deal with configuration / settings.

services/kb-mcp/src/kb_mcp/main.py
```diff
@@ -0,0 +1,82 @@
...
+        "Knowledge Base Search",
+        instructions=SERVER_CITATION_INSTRUCTIONS,
+        auth=oidc_proxy,
+        providers=[FileSystemProvider(Path(__file__).parent / "tools")],
+    )
+
+    # Bearer auth (Authorization header) — no cookies, so no allow_credentials.
+    # Wildcard + credentials would reflect any Origin (credentialed CORS hole).
+    middleware = [
+        Middleware(
```
konsti, line 55: Why do we need CORS and why is it configured with such a wide allowance?

services/kb-mcp/src/kb_mcp/settings.py
```diff
@@ -0,0 +1,22 @@
+"""kb_mcp-specific server settings (extends unique_mcp ServerSettings)."""
+
+from pydantic import Field, HttpUrl
+from unique_mcp.settings import ServerSettings
+
+
+class KbMcpServerSettings(ServerSettings):
```
konsti, line 7: Rather than this isolated setting I would keep all configuration transparently in a single module.

services/kb-mcp/src/kb_mcp/settings.py
```diff
@@ -0,0 +1,22 @@
+"""kb_mcp-specific server settings (extends unique_mcp ServerSettings)."""
+
+from pydantic import Field, HttpUrl
+from unique_mcp.settings import ServerSettings
```
konsti, line 4: Also here: this shared inheritance makes it hard to review. I would discourage it. Rather more code and more transparency then DRYness for DRYness sake.

services/kb-mcp/src/kb_mcp/scope_resolver.py
```diff
@@ -0,0 +1,78 @@
+"""Resolve knowledge-base folder/scope ids for search result deep links."""
```
konsti, line 1: Doesn't this belong into a module related to KB searches? Why is it in root?

services/kb-mcp/src/kb_mcp/tools/content_tree.py
```diff
@@ -0,0 +1,353 @@
+"""Knowledge Base content-tree tool — browse, list, and fuzzy-search visible files.
```
konsti, line 1: The tools are hard to review. Maybe I'm just not used to Python code, but it feels very clattered. Unnecessary constants like `_TOOL_DESCRIPTION` and `_META` etc.

Maybe less relevant in the age of AI, but I'm still a believer in clean / minimal code.

konsti (commented):
I've reviewed the server code, not in detail, more in structure.
I did not look at Github / Deployment and Tests.

services/kb-mcp/src/kb_mcp/main.py
```diff
@@ -0,0 +1,82 @@
...
+from kb_mcp.references import SERVER_CITATION_INSTRUCTIONS
+
+
+def main() -> None:
+    # Single .env for local (toolkit prefers unique.env; mcp/zitadel prefer their
+    # own names — loading into the process env covers all of them).
+    # override=True so service .env wins over stale shell UNIQUE_MCP_* exports.
+    load_dotenv(Path.cwd() / ".env", override=True)
+    # Opt-in via OTEL_* (e.g. OTEL_TRACES_EXPORTER=console locally).
+    configure_tracing(service_name="kb-mcp")
```
konsti, line 27: Logging & Tracing are the only two service I find useful to come from a shared package. But this package IMO belongs into this repo for AI and human review sake!

services/kb-mcp/deploy/helm-charts/kb-mcp/values.yaml
```diff
@@ -0,0 +1,192 @@
...
+#     UNIQUE_MCP_PUBLIC_BASE_URL: https://kb-mcp.<tenant>.unique.app
+#
+#     # Unique web app origin (the product UI), NOT this MCP.
+#     # Optional — only for knowledge-upload deep links in citations:
+#     # Leave unset to fall back to unique://content/{id} links.
+#     UNIQUE_MCP_FRONTEND_BASE_URL: https://<tenant>.unique.app
+#
+# Example for loading secrets (uncomment and customize as needed):
+#   envVars:
+#     - name: UNIQUE_APP_KEY
```
nicolae536, line 38: I would rather expose chart configs and map them into env vars, like that changing env var names becomes easy and it's not the concern of the deployment.

services/kb-mcp/src/kb_mcp/tools/content_tree.py
```diff
@@ -0,0 +1,353 @@
...
+    "a file; use the content_id for read_file calls.\n"
+    "Listings are cached per user (~30 min); repeat calls are fast. When the "
+    "user says they added, deleted, or changed files and needs a fresh tree, "
+    "call with refresh=true (expect a slower ~20s refetch)."
+)
+
+_META = merge_tool_meta(
+    {
+        "unique.app/icon": "folder-tree",
+        "unique.app/system-prompt": (
```
nicolae536, line 167: one thing which I'm not sure is this whole `unique.app/` metas because if I remember correctly but maybe I got it wrong @PascalHauriUnique told us that we should use the standard more and more so this would probably be in the tool description, are we moving away from this or not so that I understand where we are going ?

services/kb-mcp/src/kb_mcp/tools/content_tree.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+def _display_path_segments(segments: Sequence[str]) -> list[str]:
+    """Path segments for display and filtering (sentinel dropped, brackets stripped)."""
+    return [
+        _normalize_path_segment(s) for s in segments if s != _NO_FOLDER_PATH_SENTINEL
+    ]
+
+
+def _display_path(segments: Sequence[str]) -> str:
```
nicolae536, line 112: maybe we can move this out to a path utils - to make file shorter and easier to digest

services/kb-mcp/src/kb_mcp/tools/content_tree.py
```diff
@@ -0,0 +1,353 @@
...
+    url = file_reference_url(
+        content_info.id,
+        metadata=content_info.metadata,
+        owner_id=content_info.owner_id,
+        frontend_base_url=frontend_base_url,
+    )
+    return markdown_citation_link(display, url)
+
+
+def _get_tree_cache() -> AsyncTTLCache:
```
nicolae536, line 137: This is just a personal question not a request to change anything -> is this the standard in python I saw also agents promoting this module global vars which act like a singleton and the are used through a factory function, to me it seems weird that there is no DI container with a singleton that's why I'm asking

services/kb-mcp/src/kb_mcp/tools/read_file.py
```diff
@@ -0,0 +1,310 @@
...
+
+def _virtual_page_token_bounds(
+    start_page: int, end_page: int, max_tokens_per_call: int
+) -> tuple[int, int]:
+    token_start = (start_page - 1) * max_tokens_per_call
+    token_end = end_page * max_tokens_per_call
+    return token_start, token_end
+
+
+@tool(
```
nicolae536, line 235: I will ask just to be sure we support this case, because I have the feeling we collect all tools and register them based on the `@tool` decorator.
In other mcp's clients request different configs like we want to ingest transcripts or not in teams mcp for example which means we do not need to ship some tools, assuming we want to migrate to pythong a microsoft mcp how easy it is based on env vars or feature flags to decide if a tool is exposed or not ? Cause ideally we would not deploy and then disable tools in Unique we want to not expose those tools at all.

services/kb-mcp/src/kb_mcp/tools/content_tree.py
nicolae536: I have the same feeling -> but it seems to become a bit easier I think it's just hard to digest at the beginning the new syntax and in the AI world it takes more time since we do not write the code anymore.

services/kb-mcp/src/kb_mcp/tools/read_file.py
```diff
@@ -0,0 +1,310 @@
...
+        metadata=content.metadata,
+        frontend_base_url=KbMcpServerSettings().frontend_base_url_str(),
+    )
+    header = markdown_citation_link(content.title or content.key, url)
+    first.text = f"{header}\n\n{first.text}"
+    return result
+
+
+def _ok(text: str) -> CallToolResult:
+    return CallToolResult(content=[TextContent(type="text", text=text)])
```
nicolae536, line 114: I see we use `text` mainly as return types to llm this is more a question for me to elevate my knowledge is this bettern than returning json + descriptions ? what's the reason behind it to serialize it to text.

services/kb-mcp/src/kb_mcp/tools/read_file.py
```diff
@@ -0,0 +1,310 @@
...
+        if total_tokens <= max_tokens_per_call:
+            return _ok(_render_with_page_markers(chunks))
+        return _error(
+            f"file has ~{total_tokens} tokens across {total_pages} pages; "
+            "specify start_page/end_page to read a portion."
+        )
+
+    s = start_page if start_page is not None else 1
+    e = end_page if end_page is not None else total_pages
+    if s < 1 or s > e or s > total_pages:
```
nicolae536, line 144: now sure how relevant this is but do you need logging for this ? assume this mcp is at the client we are kindof blind for what happens in this tools if someone says this did not work, we do not have access to db or other things mainly we have logs and they are almost empty.

services/kb-mcp/src/kb_mcp/tools/search.py
```diff
@@ -0,0 +1,124 @@
...
+    config: SearchToolConfig = Depends(get_tool_config(SearchToolConfig)),
+) -> CallToolResult:
+    """Search the knowledge base using ``SearchToolConfig`` from the config meta key."""
+
+    try:
+        # In-body (not Depends) so identity-refusal ValueError surfaces as a tool error.
+        settings = await get_unique_settings_async()
+        service = KnowledgeBaseInternalSearchService.from_config(
+            config.service_config
+        ).bind_settings(settings)
```
nicolae536, line 89: this seems a funny construction of a object why not passed as argument of `from_config` ? no need to fix it I saw it comes from a package

services/kb-mcp/src/kb_mcp/scope_resolver.py
```diff
@@ -0,0 +1,78 @@
...
+            resolved[chunk.id] = scope
+        else:
+            missing.add(chunk.id)
+
+    if not missing:
+        return resolved
+
+    user_id = settings.authcontext.get_confidential_user_id()
+    company_id = settings.authcontext.get_confidential_company_id()
+    semaphore = asyncio.Semaphore(_LOOKUP_CONCURRENCY)
```
nicolae536, line 48: I'm wandering if this is at the right level cause you can call this method 2 times `resolve_scope_ids` than you have concurrency of 16 - was this your intention ? I would expect you have a semaphore per user_id+company_id combination

services/kb-mcp/src/kb_mcp/tools/content_tree.py
```diff
@@ -0,0 +1,353 @@
...
+# Statement (`.to_dict()` per call) so no shared mutable dict is handed to
+# downstream service calls.
+_DEFAULT_METADATA_FILTER_STATEMENT = Statement(
+    operator=Operator.NOT_CONTAINS,
+    path=["folderIdPath"],
+    value="user-memory",
+)
+
+
+class ContentTreeToolConfig(BaseModel):
```
nicolae536, line 52: I'm wandering if this things should not be separated or maybe use folders like 

```
tools
   search
       tool.py
       config.py       
```

it's just an idea feel free to ignore it if you think it doesn't make sense

services/kb-mcp/tests/test_content_tree_tool.py
```diff
@@ -0,0 +1,485 @@
...
+    yield
+    ct_module._tree_cache = None
+
+
+def test_match_target_matches_service_definition():
+    assert set(MatchTarget.__args__) == set(ServiceMatchTarget.__args__)
+
+
+@pytest.mark.asyncio
+async def test_mode_tree_calls_render_visible_tree_only():
```
nicolae536, line 77: this literally exposes the internal implementation - does this test bring any value besides locking the code in place ?

services/kb-mcp/tests/test_content_tree_tool.py
```diff
@@ -0,0 +1,485 @@
...
+        return_value=[(_make_content_info("c1"), ["Contracts", "a.pdf"])]
+    )
+    with patch("kb_mcp.tools.content_tree.ContentTree", return_value=mock_tree):
+        result = await content_tree(
+            mode="list",
+            config=ContentTreeToolConfig(),
+        )
+
+    mock_tree.resolve_visible_file_paths_async.assert_called_once()
+    mock_tree.render_visible_tree_async.assert_not_called()
```
nicolae536, line 105: same here why does the test care about what we call internally ? I would go more in the direction
1. I input this + mock
2. Call the method
3. I expect this output
I'm not a pro at unit testing but this tests fail every time code changes and I'm wandering what value does that bring they do not actually test that during a refactoring nothing broke

services/kb-mcp/tests/test_search_tool.py
```diff
@@ -0,0 +1,413 @@
...
+    assert len(result.content) == 2
+
+
+@pytest.mark.asyncio
+async def test_search_uses_defaults_when_no_config_provided():
+    chunks = [_make_chunk("default")]
+    mock_service = MagicMock()
+    mock_service.bind_settings.return_value = mock_service
+    mock_service.state = MagicMock()
+    mock_service.run = AsyncMock(return_value=MagicMock())
```
nicolae536, line 139: same here I'm wandering if we cannot use `input => output` method for this at least we don't verify explicit calls.

services/kb-mcp/tests/test_search_tool.py
```diff
@@ -0,0 +1,413 @@
...
+        path, "unique://content/cont_ioi3voailf7hr011zcp6b7eh"
+    )
+    assert link == (
+        "[\\[SM\\]/AlpenSys/Audit_Report_AlpenSys_FY2023.pdf]"
+        "(unique://content/cont_ioi3voailf7hr011zcp6b7eh)"
+    )
+    assert not link.startswith("[[")
+
+
+def test_chunk_to_text_content_escapes_brackets_in_title():
```
nicolae536, line 286: this test looks perfect from my perspective 
=> clear input
=> clear call
=> clear assert on output

nicolae536 (changes requested):

proschu2:
Thanks both for the detailed review — I went through every thread here and used it to drive the follow-up cleanup in [#756](https://github.com/Unique-AG/connectors/pull/756). Quick rundown of what happened to each point:

@konsti 

- Central config module / stop the scattered settings inheritance (threads on main.py:25, settings.py:4,7) → done: consolidated ServerSettings/KbMcpServerSettings/etc. into one flat kb_mcp.settings.Settings, no inheritance chain.
- Why CORS, and why so wide? (main.py:55) → agreed, it was dead weight and from tutorial:  /mcp is server-side and OAuth redirects are top-level navigation, so CORS never applied. Deleted it.
- _TOOL_DESCRIPTION/_META clutter in the tools (content_tree.py:1) → moved content_tree/read_file's descriptions into their docstrings (FastMCP infers them), dropped _TOOL_DESCRIPTION. Kept _META on search since it composes a shared constant a docstring literal can't reference, but the other two are cleaner now.
- Zitadel OIDC proxy wrapper obscuring auth in another repo (main.py:32), and logging/tracing living in the shared package (main.py:27) → I edited the `unique_mcp` version and its logging/tracing setup. Open to discussing further if you still think it's the wrong call: for me if we want to create a blueprint for MCPs it make sense to have a lib like `unique_mcp`. I am open into the discussion of moving it away from `ai` in case


@nicolae536

- Chart configs mapped to env vars (values.yaml:38) → done: mcpConfig.{app.publicBaseUrl,frontendBaseUrl,zitadel.baseUrl} now map to env vars via templates/_ext.tpl.
- Path helpers out of content_tree.py (content_tree.py:112) → done, that's tools/content_tree/path_utils.py now.
- Folder-per-tool structure (content_tree.py:52) → done exactly as suggested: tools/{search,content_tree,read_file}/{tool.py,config.py,...}.
- Logging — "we're almost blind, logs are empty" (read_file.py:144) → done: added entry/exit/error logging with a hashed correlation id across all three tools, plus a regression test asserting no raw user_id/company_id leaks into logs.
- Tests asserting internal calls instead of behavior (test_content_tree_tool.py:77,105, test_search_tool.py:139) → rewrote these to assert on the returned CallToolResult content instead of mock.assert_called_once(). Left the ones you called out as "perfect" (test_search_tool.py:286) as-is.
- The semaphore concurrency question (scope_resolver.py:48) → fixed. You were right — resolve_scope_ids was creating a fresh semaphore per call, so two concurrent search calls got lookup_concurrency each instead of sharing one budget. Went with a process-wide semaphore instead of a per-identity one — scope_lookup_concurrency now bounds total concurrent lookups across the whole process, not per call, so calling it N times concurrently no longer multiplies the load. (Considered a per-user_id+company_id semaphore first, closer to what you suggested, but it doesn't actually bound the case that matters more — many different concurrent users would still be unbounded — and adds real complexity for a feature that already degrades gracefully to unique:// links on any contention/failure. A single shared budget felt like the better tradeoff, but happy to revisit if you had a specific reason for per-identity in mind.)
- The open questions (standard unique.app/ meta usage vs description text, singleton/factory vs DI container, feature-flag-gated tool exposure, text vs JSON return types, the search.py:89 construction you said not to bother with) → these were genuine questions rather than change requests, so I left them as discussion points rather than code changes. Happy to dig into any of them if useful.

This is now up in [#756](https://github.com/Unique-AG/connectors/pull/756) if you want to take another pass.

nicolae536 (approved):

PR #720, opened 2026-07-31
build(teams-mcp,outlook-semantic-mcp,sharepoint-connector,confluence-connector,kyckr-mcp,temenos-mcp,ir-demo-mcp,ci,deps): dockerfiles and dependabot harmonized and updated

migg24:
- node `26.4.0` to `26.5.0`, build and runtime digests identical to the ones `docker-base/nestjs/26` pins
- pnpm `10.28.1` to `10.34.5`, in the images and in `packageManager` (root plus all seven services). pnpm defaults to `pmOnFail: download`, so a mismatch between the pnpm in the image and the declared `packageManager` makes pnpm fetch the declared version on every invocation, during the build and inside the chart migration hook. Keeping both on one version is what stops that.
- runtime user is the upstream `node` user (uid 1000) instead of a hand-rolled `nestjs` (uid 1001). The base Helm chart already renders `runAsUser: 1000`, so until now the process ran as a uid that owned none of its own files. Image and chart agree from here on.
- runtime hardening matches the base: `apt-get upgrade`, `procps`, npm and npx removed after pnpm is installed
- `curl` dropped. All chart probes are `httpGet`, nothing execs curl. The only curl health checks in the repo belong to the local Confluence test container.
- `PNPM_STORE_DIR` pinned to the path the base uses. The cache mounts targeted `/pnpm/store`, which pnpm never used, so dependency caching in CI has been a no-op.
- `PNPM_HOME` and the matching `PATH` prepend removed. They only serve `pnpm add -g` and `pnpm setup`, which no build performs. pnpm itself comes from `npm install -g` and is already on PATH.
- Dependabot now governs the base image updates. Docker updates run Friday 06:00 Europe/Berlin, the same slot as the `docker-base` rebuilds. Updates are grouped into a single PR because one node bump would otherwise open seven and hit `open-pull-requests-limit`. Node majors are ignored, a new node line is a deliberate `docker-base` decision. `rebase-strategy: disabled` freezes an open PR on the digest it was opened with, so a manual waiting period actually ages the artifact under review. The config file is renamed to `dependabot.yaml` for consistency with every workflow file in `.github/`. Change to 3 days cooldown.


irina-unique (approved):

services/confluence-connector/deploy/Dockerfile
```diff
@@ -64,38 +64,35 @@ RUN --mount=type=cache,target=/root/.cache/turbo \
...
-COPY --from=builder --chown=nestjs:nodejs /app/services/${APP_NAME}/dist ./dist
-COPY --from=builder --chown=nestjs:nodejs /app/out/node_modules ./node_modules
-COPY --from=builder --chown=nestjs:nodejs /app/out/package.json ./package.json
+COPY --from=builder --chown=node:node /app/services/${APP_NAME}/dist ./dist
+COPY --from=builder --chown=node:node /app/out/node_modules ./node_modules
+COPY --from=builder --chown=node:node /app/out/package.json ./package.json
 
 # Run as non-root for security
-USER nestjs
+USER node
```
nicolae536, line 95: any reason for changing the user ?

nicolae536 (approved):

services/confluence-connector/deploy/Dockerfile
migg24: yes, the base image comes preconfigured with that user, and I removed the custom user creation: 
```
&& groupadd --system --gid 1001 nodejs \
&& useradd --system --uid 1001 --gid nodejs --shell /bin/bash --create-home nestjs
```
so the nestjs user does not exist anymore, aligned with all other dockerfiles.

PR #737, opened 2026-08-03
fix(hello-mcp): run the image as uid 1000, matching the chart

migg24:
The chart pulls in the base library chart, which renders runAsNonRoot: true and runAsUser: 1000 pod and container side when values do not override it, and hello-mcp does not override it. The image created its `app` user with uid and gid 1001, so the process ran as 1000 while every file it owns, including the venv copied in with --chown=app:app, belonged to 1001. It works because the app only reads, but it is an inconsistency waiting for the first write.

uid and gid are now 1000. Nothing else changes: the base images stay upstream and digest pinned, uv stays pinned, /tmp stays the writable emptyDir the chart mounts.


proschu2 (approved):
LGTM — aligns image uid/gid with the base chart's default runAsUser: 1000. Same footgun exists on kb-mcp; mirroring there.

PR #748, opened 2026-08-06
feat(backstop-mcp): add Backstop HTTP client and OAuth credential bridge

nicolae536:
Add the shared Backstop HTTP client (JSON:API, pagination, retries) and the OAuth credential bridge.

Refs: UN-23675


klcd (commented):
Python-only pass, focused on two things: code that hand-rolls something a dependency you already ship does better, and naming.

The recurring theme is that `httpx` and `pydantic` are both in `pyproject.toml` but only half-used. `httpx.AsyncClient` is treated as a bare `request()` transport, so base-URL joining, Basic auth and URL parsing are all reimplemented by hand (three separate `base_url.rstrip('/') + path` sites). And pydantic is used for the outermost envelope but then shadowed by parallel dataclasses and by narrowing helpers in `coerce.py` that duplicate `StringConstraints`.

None of these are correctness blockers on their own — but together they are a few hundred lines that could just not exist.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+    """
+
+
+def build_auth_headers(username: str, api_token: str) -> dict[str, str]:
+    """Build the `Authorization: Basic ...` + `token: true` headers Backstop expects.
+
+    Every user connects with a personal API token (not a password), so `token: true` is
+    always sent — see https://backstopsolutions.elevio.help/en/articles/1018 and .../236.
+    """
+    basic_auth = base64.b64encode(f"{username}:{api_token}".encode()).decode()
```
klcd: `httpx` already does this: `httpx.BasicAuth(username, api_token)`. You can pass it per-request as `client.request(..., auth=...)`, which drops the base64 encoding, the two header-name constants, and this whole function.

Also worth noting `token: true` is set in two places — here and in `factory._SHARED_CLIENT_HEADERS`. Once you move to `auth=`, only the factory sets it.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+    return {_AUTHORIZATION_HEADER: f"Basic {basic_auth}", _TOKEN_HEADER: "true"}
+
+
+def is_extended_profile_path(path: str) -> bool:
+    return any(marker in path for marker in _EXTENDED_PROFILE_MARKERS)
+
+
+def build_url(base_url: str, path: str) -> str:
```
klcd: Re-anchoring my earlier comment now the stack is split.

To correct myself: `pathlib` is the wrong tool here — it's for filesystem paths and will happily mangle `https://` into `https:/`. The right one is already imported: `httpx.URL`.

Give the shared client a `base_url` (`httpx.AsyncClient(base_url=self._settings.base_url, ...)` in the factory) and httpx does relative-path merging for you, correctly, including the trailing-slash case this hand-rolls with `rstrip('/')` plus a separator. That same `rstrip('/') + path` pattern appears three times in this PR (here, `factory.verify_credential`, `provider.authorize`), which is the tell.

The host-pinning check is still worth keeping — but `httpx.URL(...).host` returns `str`, so you don't need the `.netloc` bytes plus `.decode('ascii', 'replace')` dance.

```python
def build_url(base_url: str, path: str) -> httpx.URL:
    url = httpx.URL(base_url).join(path)
    if url.host != httpx.URL(base_url).host:
        raise BackstopUntrustedUrlError(path, httpx.URL(base_url).host)
    return url
```

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+def _metric_route(path: str) -> str:
+    """Coarse, bounded label for a request path: its first segment only.
+
+    Full paths carry record ids, which would make the metric's cardinality unbounded.
+    """
+    without_query = path.split("?", 1)[0]
+    is_absolute = without_query.startswith(("http://", "https://"))
+    stripped = without_query.removeprefix("http://").removeprefix("https://")
```
klcd: Hand-rolled URL parsing again — `removeprefix('http://').removeprefix('https://')`, then a manual `split('?')`, then dropping the host segment by index. `httpx.URL` handles absolute and relative in one shot:

```python
def _metric_route(path: str) -> str:
    segments = httpx.URL(path).path.strip("/").split("/")
    return f"/{segments[0]}" if segments[0] else "/"
```

No query stripping needed (`.path` excludes it) and no is-it-absolute branch (`.path` never contains the host).

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+                max_records=max_records,
+                first_page_params=first_page_params,
+            )
+        )
+
+        if schema is None:
+            return raw_result  # pyright: ignore[reportReturnType]
+
+        try:
+            items = [TypeAdapter(schema).validate_python(item) for item in raw_result.items]
```
klcd: `TypeAdapter(schema)` is constructed **inside the comprehension**, so you build one adapter per item — on a 10,000-record pagination walk that's 10,000 schema builds. Building a `TypeAdapter` is the expensive part; validating with it is the cheap part.

Hoist it out of the loop at minimum. Better, since `_deserialize` on line 76 has the same problem on every single request, cache adapters process-wide:

```python
@functools.cache
def _adapter_for(schema: type) -> TypeAdapter[object]:
    return TypeAdapter(schema)
```

The `except ValidationError` block here is also a verbatim copy of the one in `_deserialize` — with a cached adapter both call sites can share one helper.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+    """Build the `Authorization: Basic ...` + `token: true` headers Backstop expects.
+
+    Every user connects with a personal API token (not a password), so `token: true` is
+    always sent — see https://backstopsolutions.elevio.help/en/articles/1018 and .../236.
+    """
+    basic_auth = base64.b64encode(f"{username}:{api_token}".encode()).decode()
+    return {_AUTHORIZATION_HEADER: f"Basic {basic_auth}", _TOKEN_HEADER: "true"}
+
+
+def is_extended_profile_path(path: str) -> bool:
```
klcd: Naming: `is_extended_profile_path` reads as "is this the path of an extended profile" — in a CRM codebase full of people and organizations, "profile" strongly suggests a user profile. What it actually means is "does this endpoint get the slow timeout and the bigger page size".

`is_slow_endpoint` (and `_SLOW_ENDPOINT_MARKERS`) says that. The same wording leaks into `settings.py`'s comment and into `_default_page_size`, so it's worth fixing at the source.

Separately: this and `_default_page_size` are the two places that branch on it, and both pick between a `default_*` and a `report_*` setting. A small `TimeoutProfile`/`EndpointProfile` value object returning `(timeout, page_size)` would collapse both branches into one lookup.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+        raise BackstopResponseSchemaError(path, schema.__name__, exc) from exc
+
+
+# /reports and /{entity}/{id}/analytics are the calls Backstop docs call out as legitimately
+# slow (up to ~30s per 500 records) — they get the extended timeout and the larger
+# report-sized page default; everything else gets the ordinary CRUD profile.
+_EXTENDED_PROFILE_MARKERS = ("/reports", "/analytics")
+
+
+class BackstopAuthError(Exception):
```
klcd: Re-anchoring: these two belong in `errors.py`. I read the docstring there explaining that they stay here because they aren't `ToolError`s — but "which base class it has" isn't a reason to live in a different module. `errors.py` is the errors module; a reader looking for the Backstop exception list should find all of it in one place, and the docstring can keep making the ToolError-vs-plain distinction right next to the classes it describes.

services/backstop-mcp/src/backstop_mcp/backstop_client/errors.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+class _JsonApiErrorBody(BaseModel):
+    errors: list[_JsonApiError]
+
+
+_ERROR_BODY_ADAPTER = TypeAdapter(_JsonApiErrorBody)
+
+
+@dataclass(frozen=True, slots=True)
```
klcd: `BackstopErrorDetail` is a field-for-field copy of `_JsonApiError` above it (`code`, `title`, `detail`, all `str | None`), plus a `_to_error_detail` converter on line 167 whose entire job is to copy three attributes across.

Two names and a converter for one concept. Drop the dataclass, put `message` on `_JsonApiError` as a `@property`, rename it `BackstopErrorDetail`, and delete `_to_error_detail` — `_parse_error_details` then returns the validated models directly.

That also fixes `client._errors_for_log`, which hand-builds `{"code": ..., "title": ..., "detail": ...}`: with a pydantic model it's `error.model_dump()`.

services/backstop-mcp/src/backstop_mcp/backstop_client/errors.py
```diff
@@ -0,0 +1,254 @@
...
+            return kind
+    return None
+
+
+def _parse_retry_after(response: httpx.Response) -> float | None:
+    retry_after = cast("str | None", response.headers.get("Retry-After"))
+    if retry_after is None:
+        return None
+    try:
+        return float(retry_after)
```
klcd, line 230: `float(retry_after)` only handles the delta-seconds form. Per RFC 9110 §10.2.3 `Retry-After` is *either* delta-seconds *or* an HTTP-date, and the HTTP-date form is common from CDNs and API gateways — which is what sits in front of Backstop. Today that lands in the `except ValueError` and silently degrades to exponential backoff.

The stdlib parses the date form:

```python
from email.utils import parsedate_to_datetime
...
try:
    return float(retry_after)
except ValueError:
    pass
try:
    return max(0.0, (parsedate_to_datetime(retry_after) - datetime.now(UTC)).total_seconds())
except (TypeError, ValueError):
    logger.warning(...)
    return None
```

services/backstop-mcp/src/backstop_mcp/backstop_client/errors.py
```diff
@@ -0,0 +1,254 @@
...
+def _fallback_message(response: httpx.Response) -> str:
+    logger.debug(
+        "backstop.error_body.unparseable",
+        extra={"status_code": response.status_code, "body_length": len(response.content)},
+    )
+    return f"Backstop returned status {response.status_code} with an unparseable response body"
+
+
+def _classify_limit_kind(detail: str, code: str | None) -> LimitKind | None:
+    haystack = f"{detail} {code or ''}".lower()
```
klcd: Naming: `haystack` describes the search algorithm, not the value. It's the error text you're classifying — `error_text` or `message` says that.

On the substance: `_LIMIT_KIND_KEYWORDS` plus a word-boundary regex is a keyword search over English prose from a third party, and `retry.py` treats a `None` result as non-retryable. So a Backstop copy edit ("per 60 seconds" instead of "per minute") silently turns retryable 429s into hard failures.

If the JSON:API `code` field is populated on 429s, match on that instead and keep the prose scan as a fallback — a `StrEnum` of known codes is both cheaper and stable. Worth confirming against a real 429 body before merging.

services/backstop-mcp/src/backstop_mcp/backstop_client/pagination.py
```diff
@@ -0,0 +1,106 @@
...
+
+T = TypeVar("T", default=dict[str, object])
+
+
+class _PageLinks(BaseModel):
+    next: str | None = None
+
+
+class _PageMeta(BaseModel):
+    totalResourceCount: int | None = None
```
klcd: `totalResourceCount` is a camelCase attribute on a Python model — ruff won't flag it but it reads as foreign everywhere it's used, and `PageResult` already renames it to `total_count`, so the wire spelling leaks in for exactly one line of life.

pydantic has a first-class answer:

```python
from pydantic import ConfigDict
from pydantic.alias_generators import to_camel

class _PageMeta(BaseModel):
    model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True)
    total_resource_count: int | None = None
```

Worth setting on a shared base model for every JSON:API type in this package — `json_api.py` will need the same thing as soon as an attribute payload has more than one word in its name.

services/backstop-mcp/src/backstop_mcp/backstop_client/pagination.py
```diff
@@ -0,0 +1,106 @@
...
+    (`type`, `id`) — JSON:API repeats an included resource on each page that references it.
+    """
+
+    items: list[T] = field(default_factory=list)
+    included: list[dict[str, object]] = field(default_factory=list)
+    total_count: int | None = None
+    truncated: bool = False
+
+
+@dataclass
```
klcd: `PaginationRequest` is a parameter object wrapping four arguments for a function called from exactly one place (`client.paginate`), which then has to construct it inline. It buys nothing over four keyword arguments, and it costs a class, a docstring, and an import at the call site.

Also a consistency point: `PageResult` and `PaginationRequest` are plain dataclasses in a module whose other three types are pydantic models. Pick one — for `PageResult` specifically, a generic `BaseModel` would let `paginate()` validate items into it rather than re-validating them by hand afterwards (see `client.py:260`).

services/backstop-mcp/src/backstop_mcp/backstop_client/settings.py
```diff
@@ -0,0 +1,45 @@
...
+only on the fields it actually reads.
+
+Frozen, so "the factory owns the one set of settings" is enforced rather than asserted.
+"""
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class BackstopTransportSettings:
```
klcd: These are frozen dataclasses in a project that depends on pydantic and pydantic-settings, so nothing here is validated: `base_url` is a bare `str`, timeouts and page sizes are unconstrained numbers. `BackstopConfig` in `config.py` *does* constrain them (`gt=0`, `ge=1`, `le=500`) — but those constraints are lost the moment `create_app` copies the values across into this shape, which is precisely the boundary where a hand-written translation can drop a field.

Make them `BaseModel`s with `model_config = ConfigDict(frozen=True)` and the constraints ride along:

```python
class BackstopTransportSettings(BaseModel):
    base_url: HttpUrl
    default_timeout_seconds: PositiveFloat
    report_page_size: int = Field(ge=1, le=500)
```

Same for `RetrySettings`. `HttpUrl` also gets you a parsed, normalized URL instead of a string that three different call sites `rstrip('/')`.

services/backstop-mcp/src/backstop_mcp/backstop_client/factory.py
```diff
@@ -0,0 +1,232 @@
...
+
+        Called from the login form's submit handler (see `auth/provider.py`) before minting an
+        authorization code, so it cannot go through `BackstopClient` (there is no stored
+        credential yet) — but it still uses the shared pool and configured timeout. Returns
+        True/False for a definite valid/invalid answer; raises `BackstopUnreachableError` if
+        Backstop itself couldn't be reached (network error, 5xx), which is not the same
+        failure mode as "wrong token" and should be shown to the user differently.
+        """
+        client = await self._shared_http_client()
+        url = self._settings.base_url.rstrip("/") + _VERIFICATION_PATH
```
klcd: Third copy of `base_url.rstrip('/') + path` in this PR (after `client.build_url` and `provider.authorize`). Set `base_url=` on the `AsyncClient` at line 219 and this becomes `await client.get(_VERIFICATION_PATH, ...)` — httpx does the join, correctly, in one place.

services/backstop-mcp/src/backstop_mcp/backstop_client/factory.py
```diff
@@ -0,0 +1,232 @@
...
+        async with self._http_client_lock:
+            if self._http_client is None or self._http_client.is_closed:
+                self._http_client = httpx.AsyncClient(
+                    headers=_SHARED_CLIENT_HEADERS,
+                    limits=httpx.Limits(max_connections=100, max_keepalive_connections=20),
+                )
+            return self._http_client
+
+
+def create_backstop_client_factory(
```
klcd: Re-anchoring: this is a one-line wrapper that forwards its arguments to the constructor unchanged and is used once. `BackstopClientFactory(...)` at the call site reads better and is one less name to know.

services/backstop-mcp/src/backstop_mcp/features/auth/crypto.py
```diff
@@ -0,0 +1,91 @@
...
+    ciphertext = AESGCM(key).encrypt(nonce, plaintext, None)
+    return _ENVELOPE_VERSION + nonce + ciphertext
+
+
+def decrypt_credential(blob: bytes, key: bytes) -> BackstopCredentialSecret:
+    """Decrypt a credential blob previously produced by `encrypt_credential`."""
+    if len(blob) < len(_ENVELOPE_VERSION) + _NONCE_SIZE:
+        raise InvalidCredentialEnvelopeError("Credential blob is too short to be valid")
+
+    version = blob[: len(_ENVELOPE_VERSION)]
```
klcd: This is a hand-rolled cryptographic envelope: a version byte, a nonce, and the slicing arithmetic to take it apart again (`blob[len(_ENVELOPE_VERSION) : len(_ENVELOPE_VERSION) + _NONCE_SIZE]` — three occurrences of the same offset expression).

`cryptography` already ships the format you're building: `cryptography.fernet.Fernet` is versioned, authenticated (AES-CBC + HMAC), and handles nonce generation and parsing. More to the point, `MultiFernet` gives you **key rotation** — re-keying with this envelope currently means a migration script, since the version byte identifies the format but not the key.

If you want to keep AES-GCM specifically, at least stop recomputing offsets: `memoryview` plus a couple of named slice constants, or `struct.unpack`.

Naming while here: `blob` says "bytes" and nothing else. `envelope` is the word the docstring already uses.

services/backstop-mcp/src/backstop_mcp/features/auth/login_form.py
```diff
@@ -0,0 +1,70 @@
...
+    safe_request_id = html.escape(request_id, quote=True)
+    safe_csrf_token = html.escape(csrf_token, quote=True)
+    safe_username = html.escape(username, quote=True)
+    client_label = f"{html.escape(client_name)} wants to connect to" if client_name else "Connect"
+
+    error_html = ""
+    if error:
+        error_html = f'<p class="error">{html.escape(error)}</p>'
+
+    return f"""<!DOCTYPE html>
```
klcd: A 35-line HTML document in an f-string, with CSS braces doubled up (`{{`/`}}`) to survive formatting, and five hand-written `html.escape()` calls guarding it. This is exactly what a template engine is for, and Starlette (already a transitive dependency via FastMCP) ships `Jinja2Templates`.

With Jinja2, autoescaping is on by default, so the class of bug where someone adds a sixth interpolation and forgets to escape it stops existing. It also gets the CSS out of Python.

Concretely, one of the five is already inconsistent: line 29 uses `html.escape(client_name)` without `quote=True` while the other four pass it. It happens to be safe because it lands in element text rather than an attribute — but that's a fact a reader has to verify by hand, and it's precisely the reasoning autoescaping removes.

services/backstop-mcp/src/backstop_mcp/features/auth/provider.py
```diff
@@ -0,0 +1,685 @@
...
+                select(OAuthTokenRow).where(
+                    (OAuthTokenRow.access_token_hash == token_hash)
+                    | (OAuthTokenRow.refresh_token_hash == token_hash)
+                )
+            )
+            row = result.scalar_one_or_none()
+            if row is not None:
+                row.revoked_at = datetime.now(UTC)
+
+    async def revoke_token_family_for_subject(self, subject: str) -> None:
```
klcd: Naming: `revoke_token_family_for_subject` says "family", but the query below filters on `subject` only — it revokes every live token for that subject across *all* families. `_revoke_family` on line 620 is the one that actually operates on a family, so the two names collide on the concept that distinguishes them.

`revoke_all_tokens_for_subject` is what it does. It'd also match the `CallerAuthContext` protocol, which already calls the same behaviour `revoke_current_subject_tokens` — right now the same operation has two different names depending on which layer you're standing in.

services/backstop-mcp/src/backstop_mcp/db/models.py
```diff
@@ -1,5 +1,161 @@
...
+
+    `client_metadata_json` stores the full serialized `mcp.shared.auth.OAuthClientInformationFull`
+    (it has ~15 optional RFC 7591 fields) rather than mapping each field to its own column, so
+    nothing is lost or needs re-deriving when reconstructing the client for `get_client()`.
+    """
+
+    __tablename__: str = "oauth_clients"
+
+    client_id: Mapped[str] = mapped_column(String, primary_key=True)
+    client_metadata_json: Mapped[str] = mapped_column(Text)
```
klcd: Storing a pydantic model as a `Text` column means every read and write has to remember to serialize: `model_dump_json()` in `register_client`, `model_validate_json()` in both `get_client` and `handle_login_get`. Three call sites that must agree, with the type system unable to help — the column is just `str`.

SQLAlchemy's `TypeDecorator` moves that into the mapping, so the attribute is typed as the model and the (de)serialization happens once:

```python
class PydanticJson(TypeDecorator[T]):
    impl = JSONB
    def process_bind_param(self, value, dialect): return value.model_dump(mode="json")
    def process_result_value(self, value, dialect): return self.model_type.model_validate(value)

client_metadata: Mapped[OAuthClientInformationFull] = mapped_column(
    PydanticJson(OAuthClientInformationFull)
)
```

Also worth it for `CustomFieldSchemaSnapshot.payload`, which is currently typed `Mapped[object]`. And `JSONB` over `Text`/`JSON` on Postgres, so these are queryable.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Done in 69071378 — per-request `httpx.BasicAuth`, dropped `build_auth_headers`; `token: true` stays only on the shared client headers.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Done in 69071378 — shared `AsyncClient(base_url=...)`, relative paths pass through, absolute `links.next` pinned via `httpx.URL` in `resolve_request_url`.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Done in 69071378 — `metric_route` now uses `httpx.URL(path).path` (and still collapses id segments for cardinality).

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Done in 69071378 — `TypeAdapter(schema)` is built once per `paginate` call (outside the item loop). Skipped process-wide cache; typed verbs require an explicit schema, and status-only probes use `raw_request`.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Done in 69071378 — renamed to `is_slow_endpoint` / `_SLOW_ENDPOINT_MARKERS`. Left the timeout vs page-size branches as two lookups for now (no `EndpointProfile` yet).

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
nicolae536: Done — `BackstopAuthError` / `BackstopUnreachableError` live in `errors.py`.

services/backstop-mcp/src/backstop_mcp/backstop_client/errors.py
nicolae536: Done in 69071378 — single pydantic `BackstopErrorDetail` with `message`; logging uses `model_dump()`.

services/backstop-mcp/src/backstop_mcp/backstop_client/errors.py
nicolae536: Done in 69071378 — HTTP-date via `email.utils.parsedate_to_datetime`, with a 1s floor in the retry wait strategy for past/zero values.

services/backstop-mcp/src/backstop_mcp/backstop_client/errors.py
nicolae536: Done — renamed to `error_text`.

services/backstop-mcp/src/backstop_mcp/backstop_client/pagination.py
nicolae536: Done in 69071378 — `total_resource_count` with alias `totalResourceCount`.

services/backstop-mcp/src/backstop_mcp/backstop_client/pagination.py
nicolae536: Done in 69071378 — dropped `PaginationRequest`; `paginate_all` takes kwargs.

services/backstop-mcp/src/backstop_mcp/backstop_client/settings.py
nicolae536: Done in 69071378 — frozen pydantic models with Field constraints.

services/backstop-mcp/src/backstop_mcp/backstop_client/factory.py
nicolae536: Done in 69071378 — factory builds `AsyncClient(base_url=...)`; verify uses `raw_request`.

services/backstop-mcp/src/backstop_mcp/backstop_client/factory.py
nicolae536: Done in 69071378 — dropped `create_backstop_client_factory`; `create_app` constructs `BackstopClientFactory` directly.

services/backstop-mcp/src/backstop_mcp/features/auth/provider.py
nicolae536: Done in 69071378 — renamed to `revoke_all_tokens_for_subject`.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,363 @@
...
+
+    Absolute URLs arrive from `links.next` while walking a pagination chain. They are pinned
+    to the configured host: an authenticated client following an arbitrary upstream-supplied
+    origin would send `Authorization: Basic ...` wherever that origin points.
+    """
+    if not path.startswith(("http://", "https://")):
+        separator = "" if path.startswith("/") else "/"
+        return base_url.rstrip("/") + separator + path
+
+    expected = httpx.URL(base_url).netloc
```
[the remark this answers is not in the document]
nicolae536: Done in 453f646d — `resolve_request_url` now requires matching scheme and netloc (blocks same-host http downgrade).

services/backstop-mcp/src/backstop_mcp/features/auth/provider.py
```diff
@@ -0,0 +1,685 @@
...
+            redirect_uri=AnyUrl(row.redirect_uri),
+            redirect_uri_provided_explicitly=row.redirect_uri_provided_explicitly,
+            resource=row.resource,
+            subject=row.subject,
+        )
+
+    @override
+    async def exchange_authorization_code(
+        self, client: OAuthClientInformationFull, authorization_code: AuthorizationCode
+    ) -> OAuthTokenResponse:
```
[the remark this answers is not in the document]
nicolae536: Done in 453f646d — `load_authorization_code` returns `None` when `expires_at` is in the past.

services/backstop-mcp/src/backstop_mcp/backstop_client/client.py
```diff
@@ -0,0 +1,269 @@
...
+
+            BACKSTOP_REQUESTS.add(
+                1, {"method": method, "route": route, "status": response.status_code}
+            )
+            if response.status_code == 401:
+                if self._on_auth_failure is not None:
+                    await self._on_auth_failure()
+                raise BackstopAuthError(
+                    "Backstop rejected the stored credential — please reconnect."
+                )
```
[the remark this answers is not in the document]
nicolae536: Done in 453f646d — auth-failure hook errors are logged; 401 always raises `BackstopAuthError`.

services/backstop-mcp/src/backstop_mcp/features/auth/crypto.py
nicolae536: Done in 89f5e2ba — switched credential envelopes to Fernet (unreleased, no migration).

services/backstop-mcp/src/backstop_mcp/features/auth/login_form.py
nicolae536: Done in 89f5e2ba — Jinja2 `login.html` with autoescape; `jinja2` declared in pyproject.

services/backstop-mcp/src/backstop_mcp/db/models.py
nicolae536: Done in 89f5e2ba — `oauth_clients.client_metadata` is JSONB; provider uses `model_dump(mode="json")` / `model_validate`.

services/backstop-mcp/src/backstop_mcp/backstop_client/factory.py
```diff
@@ -0,0 +1,230 @@
...
+                return False
+            raise BackstopUnreachableError(
+                f"Backstop returned unexpected status {exc.status_code} "
+                + "while verifying credentials"
+            ) from exc
+        except httpx.RequestError as exc:
+            raise BackstopUnreachableError(
+                f"Could not reach Backstop at {self._settings.base_url}"
+            ) from exc
+        return True
```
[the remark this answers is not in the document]
nicolae536: The traceback half is fixed in 8edc0281 — `raw_request` now catches `BackstopAuthError` ahead of the catch-all and logs it at info without a stack, since a rejected credential is an ordinary outcome on the login path rather than a fault. The status-only `raw_request` change you can see above already covered the `ValidationError` half.

services/backstop-mcp/pyproject.toml
```diff
@@ -36,6 +37,9 @@ packages = ["src/backstop_mcp"]
 [tool.hatch.build.targets.wheel.sources]
 "src/backstop_mcp" = "backstop_mcp"
 
+[tool.hatch.build.targets.wheel.force-include]
+"src/backstop_mcp/features/auth/templates" = "backstop_mcp/features/auth/templates"
```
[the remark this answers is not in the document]
nicolae536: Fixed in 8edc0281 — dropped the `[tool.hatch.build.targets.wheel.force-include]` block; `packages = ["src/backstop_mcp"]` already ships the templates.

Verified rather than assumed: `uv build --wheel` now succeeds and `login.html` appears exactly once in the archive.

services/backstop-mcp/src/backstop_mcp/backstop_client/pagination.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    result = PageResult()
+    seen_included: set[tuple[str, str]] = set()
+    path: str | None = first_path
+    params = first_page_params
+
+    while path is not None:
+        response = await fetch_page(path, params)
+        params = None
+        page = _PAGE_ADAPTER.validate_json(response.content)
```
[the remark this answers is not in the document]
nicolae536: Done in 374c8e28 — `paginate_all` now validates page envelopes via `deserialize`, so malformed 200 bodies raise `BackstopResponseSchemaError` (same typed failure as item validation in `paginate`).

services/backstop-mcp/src/backstop_mcp/features/auth/login_csrf.py
```diff
@@ -0,0 +1,88 @@
...
+        path=path,
+        httponly=True,
+        secure=secure,
+        samesite="lax",
+    )
+
+
+def clear_csrf_cookie(response: Response, request_id: str, *, path: str) -> None:
+    """Drop the cookie once its pending authorization has been consumed."""
+    response.delete_cookie(csrf_cookie_name(request_id), path=path)
```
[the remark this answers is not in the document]
nicolae536: Fixed in 6247348d — `clear_csrf_cookie` now takes `secure` and passes matching `secure` / `httponly` / `samesite` to `delete_cookie`, and the login success path passes `self._secure_cookies`. Asserted in `test_a_successful_login_clears_the_cookie`.

services/backstop-mcp/src/backstop_mcp/app.py
```diff
@@ -9,48 +9,113 @@
...
-        try:
-            yield
-        finally:
-            await engine.dispose()
+        async with cleanup_lifespan(session_factory, auth_config):
+            try:
+                yield
+            finally:
+                await reset_services()
+                await engine.dispose()
```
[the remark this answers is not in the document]
nicolae536: Fixed in a1bdb95c (and propagated through the stack) — `engine.dispose()` / `reset_services()` now run in an outer `finally` after `cleanup_lifespan` cancels and awaits the auth sweep, so the pool stays alive until the background task has stopped.

vabatta (approved):

PR #749, opened 2026-08-06
feat(backstop-mcp): add party ID resolver library

nicolae536:
Add the party ID resolver (name, email, trusted Party ID) used by later tools.

Refs: UN-23676


services/backstop-mcp/src/backstop_mcp/features/resolution.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+@dataclass(frozen=True)
+class NotFound:
+    query: str
+    scope: str
+    status: Literal["not_found"] = "not_found"
+
+
+type Resolution[T] = Resolved[T] | Ambiguous[T] | NotFound
```
klcd: `Resolved` / `Ambiguous` / `NotFound` are frozen dataclasses each carrying a `status: Literal[...]` field, unioned here. That is the textbook description of a pydantic **discriminated union**, and writing it as dataclasses means you get none of what pydantic would give you: no validation, no serialization, and no generated JSON schema — which is why the second half of this file exists at all.

```python
type Resolution[T] = Annotated[
    Resolved[T] | Ambiguous[T] | NotFound, Field(discriminator="status")
]
```

If you want to keep the dataclass ergonomics, `pydantic.dataclasses.dataclass` is a drop-in for the decorator and gives you `TypeAdapter(Resolution[T]).dump_python(...)` for free.

The payoff is lines 259-365: `CandidateEcho`, `AmbiguousResponse`, `NotFoundResponse`, `BatchUnresolvedEcho`, `BatchResolvedEcho` and `BatchAmbiguousResponse` are a second model of the same six concepts, and `unresolved_response` / `batch_ambiguous_response` are hand-written mappers between the two. `batch_ambiguous_response` needs **five** parameters, three of which are model classes passed in so it can instantiate them. That signature is the design telling you the two hierarchies want to be one.

services/backstop-mcp/src/backstop_mcp/features/resolution.py
```diff
@@ -0,0 +1,365 @@
...
+def _unique_labels[T](
+    candidates: Sequence[Candidate[T]],
+) -> tuple[list[str], dict[str, Candidate[T]]]:
+    """Build a collision-free label list, since an elicit enum needs distinct strings."""
+    by_label: dict[str, Candidate[T]] = {}
+    labels: list[str] = []
+    for candidate in candidates:
+        display = candidate.label
+        if display in by_label:
+            display = f"{candidate.label} [{candidate.key}]"
```
klcd: Hand-rolled label disambiguation: try the label, then `label [key]`, then `label [key] #2`, `#3`, … in a `while` loop, tracked by two parallel structures (`labels` list and `by_label` dict) that must stay in step — which the `assert` on line 216 then checks at runtime.

Since `Candidate.key` is already unique (it's the resource id), the collision loop can never run past the first fallback: two candidates sharing a label cannot also share a key. So the `while` and the `suffix` counter are unreachable. That leaves:

```python
def _unique_labels[T](candidates: Sequence[Candidate[T]]) -> dict[str, Candidate[T]]:
    seen = Counter(c.label for c in candidates)
    return {
        (c.label if seen[c.label] == 1 else f"{c.label} [{c.key}]"): c
        for c in candidates
    }
```

One return value instead of two (the caller can use `list(by_label)` for the enum), no assert needed, and it also fixes a small wart in the current version: the *first* of two colliding candidates keeps the bare label while the second gets the key suffix, so the user is asked to choose between "Capstone" and "Capstone [1234]" rather than between two equally-qualified options.

services/backstop-mcp/src/backstop_mcp/features/resolution.py
```diff
@@ -0,0 +1,365 @@
...
+        return Resolved(value=chosen.value)
+
+    assert isinstance(result, (DeclinedElicitation, CancelledElicitation))
+    return ambiguous
+
+
+# --- LLM-facing response models -------------------------------------------------------------
+
+
+class CandidateEcho(BaseModel):
```
klcd: Naming: "Echo" as a type suffix (`CandidateEcho`, `BatchResolvedEcho`, `BatchUnresolvedEcho`, and `to_echo`) isn't standard vocabulary and doesn't survive being read cold — an echo is a thing that repeats back, but these are the LLM-facing serialization of a candidate.

The module docstring already establishes the right words: these are the *response* shapes. `CandidateResponse` / `BatchResolvedResponse` matches `AmbiguousResponse` and `NotFoundResponse`, which sit in the same block and are named consistently. As it stands, six types in one section use two different naming schemes for one concept.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
```diff
@@ -0,0 +1,100 @@
...
+
+    When `party_id` is set, optional `name` is passed through on the trusted-id short-circuit
+    (no existence check). `name` is ignored when `search` is set.
+    """
+
+    party_id: str | None = None
+    search: str | None = None
+    name: str | None = None
+
+    def __post_init__(self) -> None:
```
klcd: `object.__setattr__` to mutate a frozen dataclass from `__post_init__` is a workaround for not using pydantic — and the tool you'd need is declared in this same file on line 19.

```python
class PartyResolveItem(BaseModel):

    party_id: _StrippedStr | None = None
    search: _StrippedStr | None = None
    name: str | None = None

    @model_validator(mode="after")
    def _exactly_one_selector(self) -> "PartyResolveItem":
        if (self.party_id is None) == (self.search is None):
            raise ValueError("Exactly one of party_id or search must be provided")
        return self
```

`_StrippedStr` does the stripping (add `min_length=1` and empty-to-`None` coercion comes with it), so the two `object.__setattr__` calls and the manual `or ""` / `.strip() or None` dance both disappear.

The broader point: this file has one pydantic model and three frozen dataclasses, with a docstring on line 25 explaining the split as "pydantic because it's deserialized off the wire". But `PartyResolveItem` is *also* deserialized off the wire — it's tool input from the model.

klcd (commented):
Python-only pass, same lens as my comments on #748 and #716: code that hand-rolls what a dependency already in `pyproject.toml` does, plus naming.

The dominant issue here is that pydantic is used in about half the places it should be. `resolution.py` defines ten frozen dataclasses with `status: Literal[...]` discriminators — that is a pydantic discriminated union, written out longhand — and then defines a **second**, parallel set of seven pydantic models to put the same data on the wire, plus two converter functions that take three or five `type[...]` arguments to bridge them. `types.py` has the same split inside a single file, to the point of using `object.__setattr__` to work around frozen dataclasses in order to do string-stripping that the `_StrippedStr` alias declared 60 lines above already does.

The naming issue is mostly the entity-type vocabulary: `EntityType` (StrEnum), `SearchType` (Literal), `PARTY_SEARCH_TYPES` (tuple of strings) and `_ALIASES` are four representations of one list, kept in agreement by an import-time `assert`.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
```diff
@@ -0,0 +1,100 @@
...
+@dataclass(frozen=True)
+class QuickSearchOptions:
+    """Caller-overridable knobs for `GET /quick-search`.
+
+    `full_email_match=None` means auto: true when the search looks like an email, else omit/false.
+    `filter_type=None` means omit the param (API default). Pagination is first page only;
+    `page[limit]` is aligned with `limit` by the search layer.
+    """
+
+    limit: int = 10
```
klcd: `limit: int = 10` is unconstrained, so the invariant is enforced 40 lines away in another module — `search.py:113` opens `_quick_search_params` with `assert options.limit > 0`.

As a pydantic model that's `limit: int = Field(default=10, gt=0)` and the assert goes away, along with the question of what an unvalidated `limit=0` would have done to `page[limit]`. Probably worth an upper bound too, since it's forwarded straight to `page[limit]`.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/email.py
```diff
@@ -0,0 +1,4 @@
+def looks_like_email(value: str) -> bool:
```
klcd: A whole module for a four-line predicate that `"a@@b"` and `"@ @"` both pass. Two options depending on how strict you want to be:

- pydantic's own: `from pydantic import validate_email` (or an `EmailStr` field), which is RFC-aware. Needs the `email-validator` extra.
- stdlib, no new dependency: `email.utils.parseaddr`.

Either way, this is used for exactly one decision — whether to set `filter[fullEmailMatch][eq]` in `_quick_search_params` — so it belongs next to that decision in `search.py` rather than in its own file. Naming is fine (`looks_like_email` is honest about being a heuristic); it's the file that's overkill.

services/backstop-mcp/src/backstop_mcp/features/entity_types.py
```diff
@@ -0,0 +1,71 @@
...
+    "organizations",
+    "contacts",
+    "people",
+    "employees",
+)
+
+# Backstop's `custom-field-definitions.entityType` is singular ("organization") while the API
+# path segment is plural ("organizations"). Anything not listed passes through lowercased —
+# better to index an unrecognized entity type under its own name than to drop its fields.
+_ALIASES: dict[str, EntityType] = {
```
klcd: Half of this table is redundant: every plural key maps to the enum member whose value is that same string, which is what `EntityType(collapsed)` already does. Only the six singular→plural rows carry information.

```python
_SINGULAR_ALIASES = {
    "organization": EntityType.ORGANIZATIONS,
    "contact": EntityType.CONTACTS,
    "person": EntityType.PEOPLE,
    "employee": EntityType.EMPLOYEES,
    "opportunity": EntityType.OPPORTUNITIES,
    "account": EntityType.ACCOUNTS,
}
```

Then `normalize_entity_type` tries `_SINGULAR_ALIASES.get(...)` and falls back to `EntityType(collapsed)` in a `try`. Twelve hand-maintained rows become six, and adding an enum member no longer requires remembering to add its identity row here.

(If you'd rather not maintain the singulars either, `inflection.pluralize` handles all six correctly — but a six-entry dict is probably not worth a dependency.)

services/backstop-mcp/src/backstop_mcp/features/entity_types.py
```diff
@@ -0,0 +1,71 @@
...
+KNOWN_ENTITY_TYPES: tuple[EntityType, ...] = tuple(EntityType)
+
+assert set(PARTY_SEARCH_TYPES) <= {member.value for member in EntityType}, (
+    "every SearchType must be a KNOWN_ENTITY_TYPES member"
+)
+
+
+def normalize_entity_type(entity_type: str) -> str:
+    """Normalize a CRM `entityType` or API path segment to the canonical plural form."""
+    collapsed = entity_type.strip().lower().replace(" ", "").replace("_", "").replace("-", "")
```
klcd: Two things on this line.

The `.replace(" ", "").replace("_", "").replace("-", "")` chain builds three intermediate strings to do one character-class removal. `re.sub(r"[\s_-]+", "", entity_type.strip().lower())` is one pass and also catches tabs and non-breaking spaces, which the current version misses (`"organization\u00a0"` survives `.strip()` in some locales and then fails to match).

More importantly the signature lies: it's called `normalize_entity_type` and returns `str`, but on the unknown-alias path it returns `entity_type.strip().lower()` — a value that is explicitly *not* normalized to the canonical vocabulary. Callers can't tell the two cases apart from the return type. `EntityType | None` (or `EntityType | str` if passthrough is genuinely wanted) makes the two outcomes visible, and lets `is_party_search_type` below drop its string comparison in favour of set membership on enum members.

services/backstop-mcp/src/backstop_mcp/features/entity_types.py
```diff
@@ -0,0 +1,71 @@
...
+    "opportunities": EntityType.OPPORTUNITIES,
+    "account": EntityType.ACCOUNTS,
+    "accounts": EntityType.ACCOUNTS,
+}
+
+# The entity types this connector surfaces glossaries and tools for — derived from the enum
+# rather than restated, so adding a member can't leave the two lists disagreeing.
+KNOWN_ENTITY_TYPES: tuple[EntityType, ...] = tuple(EntityType)
+
+assert set(PARTY_SEARCH_TYPES) <= {member.value for member in EntityType}, (
```
klcd: An import-time `assert` reconciling two hand-maintained lists is a sign the lists shouldn't both exist. Right now one vocabulary has four representations: `EntityType` (StrEnum), `SearchType` (Literal of four strings), `PARTY_SEARCH_TYPES` (tuple of the same four strings again), and the plural half of `_ALIASES`.

Derive instead of restate:

```python
PARTY_SEARCH_TYPES: Final = frozenset({
    EntityType.ORGANIZATIONS, EntityType.CONTACTS,
    EntityType.PEOPLE, EntityType.EMPLOYEES,
})
```

and the assert becomes structurally impossible to violate. `SearchType` as a `Literal` is the one that's hard to derive — but if it exists only to give the tool schema a closed enum, FastMCP/pydantic generate exactly that from a `StrEnum` subset too.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/search.py
```diff
@@ -0,0 +1,176 @@
...
+    else:
+        params["filter[fullEmailMatch][eq]"] = _bool_param(options.full_email_match)
+
+    if options.filter_type is not None:
+        params["filter[filterType][eq]"] = options.filter_type
+
+    return params
+
+
+def _bool_param(value: bool) -> str:
```
klcd: httpx already serializes bool query params as `true`/`false` — that's been its behaviour since 0.23, and you're on 0.28. So `_bool_param` and its three call sites can just pass the bool through.

Worth a quick check against a real request first (`filter[showAll][eq]` etc. go through `BackstopClient.get(params=...)` → `httpx.AsyncClient.request(params=...)`, so the rendering is httpx's), but if it holds, that's a function and three wrappers gone.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/search.py
```diff
@@ -0,0 +1,176 @@
...
+    return "true" if value else "false"
+
+
+def _merge_candidates(
+    documents: Sequence[_PartyCollectionDocument], *, search_type: SearchType
+) -> tuple[PartyCandidate, ...]:
+    seen_ids: set[str] = set()
```
klcd: Manual first-wins dedup with a parallel `seen_ids` set and `merged` list. A dict comprehension does it in one expression, since dicts have preserved insertion order since 3.7 and `Candidate.key` is the id you're deduping on:

```python
def _merge_candidates(documents, *, search_type):
    merged: dict[str, PartyCandidate] = {}
    for document in documents:
        for candidate in _candidates_from_document(document, search_type=search_type):
            merged.setdefault(candidate.key, candidate)
    return tuple(merged.values())
```

Same semantics, one structure instead of two that have to be kept in agreement.

Minor naming point on the same function: the parameter is `documents` but the caller passes a variable named `responses` (line 52). They're `BackstopApiCollectionDocument`s, so `documents` is the right word — rename at the call site.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/resolve.py
```diff
@@ -0,0 +1,156 @@
...
+    elicit_choice,
+    from_candidates,
+)
+
+
+def _normalize_party_id_or_search(
+    party_id: str | None,
+    search: str | None,
+) -> tuple[str | None, str | None]:
+    normalized_party_id = (party_id or "").strip() or None
```
klcd: `_normalize_party_id_or_search` re-implements what `PartyResolveItem.__post_init__` already does — strip, empty-to-`None`, and the exactly-one-of check — and the two have **already drifted**: this version additionally rejects a `party_id` containing `/`, and `PartyResolveItem` doesn't.

That matters because the two entry points take different routes. `resolve_party` (line 101) normalizes through this function, while `resolve_parties` (line 137) passes `item.party_id` straight from `PartyResolveItem` into `_resolve_one` and never sees the `/` guard.

To be clear about the blast radius: this is **not** exploitable today. Every consumer interpolates through `quote(party.id, safe='')` — `fetch_party_name` here, and `get_person`/`get_organization` over in #751 — so a `/` is percent-encoded before it reaches a path. But the guard's own comment justifies itself as protection against exactly that interpolation, which means the defence-in-depth is inconsistent depending on which entry point you came through, and nothing in the types says so.

One pydantic model used by both paths fixes it structurally: `resolve_party` takes a `PartyResolveItem` (or the same validator runs on its arguments), and there's one definition of what a valid selector is.

Smaller note: `_resolve_one` then re-asserts the same invariant on line 57. Three places expressing one rule.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
```diff
@@ -0,0 +1,100 @@
...
+        composed = " ".join(part for part in (self.first_name, self.last_name) if part)
+        return composed or None
+
+
+@dataclass(frozen=True)
+class ResolvedParty:
+    """A party identity after successful resolution."""
+
+    id: str
+    type: SearchType
```
klcd: Naming: `type` shadows the builtin, and — worse here — it collides with a *different* concept that's live in the same code. `resource.type` in `search.py:169` is the JSON:API resource type off the wire; `ResolvedParty.type` is the collection that was searched. Line 175 of that file constructs one from the other (`type=search_type`, right after reading `resource.type` into `resource_type`), so both meanings are in scope three lines apart.

`search_type` is the name the rest of the module already uses for this value — it's the parameter name on `search_by_email`, `quick_search`, `fetch_party_name` and `_candidate_from_resource`. So this is the one place the concept is called something else.

Same rename applies to `ResolvedPartyEcho.type` in `responses.py:42`, which mirrors it onto the wire — worth doing together while there are no external consumers pinned to the current field name.

klcd (commented):
Two more from a second pass over the same files — one naming issue that spans three types, and one duplicated validation that has already drifted.

services/backstop-mcp/src/backstop_mcp/features/resolution.py
nicolae536: Done in a787df2e — `Resolved` / `Ambiguous` / `NotFound` (and batch counterparts) are frozen pydantic models with a `status` discriminator.

services/backstop-mcp/src/backstop_mcp/features/resolution.py
nicolae536: Done in a787df2e — `Counter` + equal `label [key]` qualification for collisions (no `#n` loop).

services/backstop-mcp/src/backstop_mcp/features/resolution.py
nicolae536: Done in a787df2e — renamed to `*Response` / `to_candidate` / `to_resolved`.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
nicolae536: Done in a787df2e — frozen pydantic `PartyResolveItem` with `model_validator` for exactly-one-of + `/` rejection.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
nicolae536: Done in a787df2e — `QuickSearchOptions.limit` is `Field(gt=0, le=500)`; dropped the search-layer assert.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/email.py
nicolae536: Done in a787df2e — inlined `looks_like_email` next to `_quick_search_params` in `search.py`; kept the heuristic (no `email-validator` dep).

services/backstop-mcp/src/backstop_mcp/features/entity_types.py
nicolae536: Done in a787df2e — singular-only alias map; plurals via `EntityType(collapsed)`.

services/backstop-mcp/src/backstop_mcp/features/entity_types.py
nicolae536: Done in a787df2e — `re.sub` collapse; `normalize_entity_type` returns `EntityType | None`.

services/backstop-mcp/src/backstop_mcp/features/entity_types.py
nicolae536: Done in a787df2e — `PARTY_SEARCH_TYPES` derived as a frozenset of `EntityType` members; dropped the identity assert.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/search.py
nicolae536: Done in a787df2e — bools pass through to httpx; deleted `_bool_param`.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/search.py
nicolae536: Done in a787df2e — `dict.setdefault` merge; call-site renamed to `documents`.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/resolve.py
nicolae536: Done in a787df2e — single `PartyResolveItem` path for `resolve_party` and batch; dropped `_normalize_*` / redundant asserts.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
nicolae536: Done in a787df2e — renamed to `search_type` on `ResolvedParty` and the wire response.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
```diff
@@ -0,0 +1,126 @@
...
+        return None
+    return value
+
+
+_StrippedStr = Annotated[str, StringConstraints(strip_whitespace=True)]
+_NonEmptyStr = Annotated[
+    str,
+    BeforeValidator(_blank_to_none),
+    StringConstraints(strip_whitespace=True, min_length=1),
+]
```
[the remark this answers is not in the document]
nicolae536: Fixed in 438c0605 — blank selectors are coerced to `None` via a `field_validator(mode="before")` on the optional union fields (Annotated[str]-only BeforeValidator was rejecting `None`).

services/backstop-mcp/src/backstop_mcp/features/entity_types.py
```diff
@@ -0,0 +1,71 @@
...
+    try:
+        return EntityType(collapsed)
+    except ValueError:
+        return None
+
+
+def is_party_search_type(entity_type: str) -> TypeGuard[SearchType]:
+    """Whether `entity_type` normalizes to a party-searchable resource type."""
+    normalized = normalize_entity_type(entity_type)
+    return normalized is not None and normalized in PARTY_SEARCH_TYPES
```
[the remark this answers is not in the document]
nicolae536: Done in $(git rev-parse --short HEAD) — dropped the `TypeGuard[SearchType]` on the raw input; `party_search_type` returns the canonical plural `SearchType` (or None), and `is_party_search_type` is a plain bool.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/resolve.py
```diff
@@ -0,0 +1,156 @@
...
+                client,
+                search_type=search_type,
+                party_id=item.party_id,
+                search=item.search,
+                name=item.name,
+                confirm_name=confirm_name,
+                quick_search_options=quick_search_options,
+            )
+            for item in items
+        )
```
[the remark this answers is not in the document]
nicolae536: Fixed in fbab3c2c — the `/` rejection moved onto `PartyResolveItem` itself, in the `_exactly_one_selector` `model_validator(mode="after")`, so both entry points get it: `resolve_party` constructs the item, and `resolve_parties` only accepts already-validated (and frozen) `PartyResolveItem`s. There's no path into `_resolve_one` that skips it. Covered by `test_rejects_party_id_containing_slash`.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
```diff
@@ -0,0 +1,100 @@
...
+    search: str | None = None
+    name: str | None = None
+
+    def __post_init__(self) -> None:
+        party_id = (self.party_id or "").strip() or None
+        search = (self.search or "").strip() or None
+        object.__setattr__(self, "party_id", party_id)
+        object.__setattr__(self, "search", search)
+        if (party_id is None) == (search is None):
+            raise ValueError("Exactly one of party_id or search must be provided")
```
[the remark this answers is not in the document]
nicolae536: Fixed in fbab3c2c — same change as the earlier instance of this: the `/` rejection lives on `PartyResolveItem._exactly_one_selector` rather than in `resolve_party`, so the batch path can't receive an unvalidated item. Worth noting the residual test gap: traversal is exercised through `resolve_party` (`test_rejects_party_id_containing_slash`) but not directly through `resolve_parties`, so a refactor that moved the check back down would regress the batch path silently.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/email.py
nicolae536: Updated in 17903529 — `looks_like_email` now uses pydantic `validate_email` (declared `pydantic[email]`), so `a@@b` / `@ @` no longer pass.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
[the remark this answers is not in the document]
nicolae536: Fixed in fdf6894a — blank/whitespace `name` now goes through the same blank→None coercion as the selectors, so `confirm_name` still fetches.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/search.py
```diff
@@ -0,0 +1,171 @@
...
+    # schema validation; a blank `type` still falls back to "resource" here since that's a
+    # caller-side display concern, not a structural defect the schema should reject.
+    resource_type = resource.type or "resource"
+    name = resource.attributes.display_name()
+    label = name if name is not None else f"{resource_type} #{resource.id}"
+    return Candidate(
+        key=resource.id,
+        label=label,
+        value=ResolvedParty(id=resource.id, search_type=search_type, name=name),
+    )
```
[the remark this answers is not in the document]
nicolae536: Fixed in fdf6894a — `_candidate_from_resource` now stamps `ResolvedParty.search_type` from the resource's own type via `party_search_type`, falling back to the requested scope only when the wire type isn't a party SearchType. Covered by `test_enhanced_quick_search_preserves_resource_search_type`.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/types.py
```diff
@@ -0,0 +1,125 @@
...
+    party_id: _NonEmptyStr | None = None
+    search: _NonEmptyStr | None = None
+    name: str | None = None
+
+    @field_validator("party_id", "search", mode="before")
+    @classmethod
+    def _blank_selector_to_none(cls, value: object) -> object:
+        if isinstance(value, str) and not value.strip():
+            return None
+        return value
```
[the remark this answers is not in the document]
nicolae536: Fixed in fdf6894a — blank/whitespace `name` now shares the blank→None coercion with the selectors, so `confirm_name` still fetches. Covered by `test_confirm_name_fetches_when_name_is_blank`.

services/backstop-mcp/src/backstop_mcp/features/party_resolver/search.py
```diff
@@ -0,0 +1,196 @@
...
+    # collection. Fall back to the requested scope only when the wire type isn't a party
+    # SearchType (shouldn't happen for `/quick-search` hits, but keep a usable candidate).
+    resource_type = resource.type or "resource"
+    resolved_search_type = party_search_type(resource_type) or search_type
+    name = resource.attributes.display_name()
+    label = name if name is not None else f"{resource_type} #{resource.id}"
+    return Candidate(
+        key=resource.id,
+        label=label,
+        value=ResolvedParty(id=resource.id, search_type=resolved_search_type, name=name),
```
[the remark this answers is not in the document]
nicolae536: Fixed: elicit `Candidate.key` is now `{search_type}:{party_id}` so cross-type enhance_search_types hits that share an id stay distinct options.

vabatta (approved):

PR #750, opened 2026-08-06
feat(backstop-mcp): add custom-field schema discovery library

nicolae536:
Add custom-field schema discovery (fetch, cache, name resolution). Tools and the on-demand catalog land in later PRs.

Refs: UN-23677


services/backstop-mcp/src/backstop_mcp/features/custom_fields/types.py
```diff
@@ -0,0 +1,76 @@
...
+
+@dataclass(frozen=True)
+class AllowedValue:
+    """One picklist / LOV entry, usable to validate a write before attempting it."""
+
+    id: str | None
+    label: str
+
+
+@dataclass(frozen=True)
```
klcd: `CustomFieldDefinition` (and `AllowedValue`) being frozen dataclasses rather than pydantic models is the root of the three-way duplication in this PR. The same fields are re-declared as `_DefinitionPayload`/`_AllowedValuePayload` in `snapshot.py:27` and again as `CustomFieldDefinitionEcho`/`AllowedValueEcho` in `responses.py:28` — with the *same names* in all three — because a dataclass can't serialize or validate itself.

```python
class CustomFieldDefinition(BaseModel):
    ...
```

That one change collapses all three copiers below (`dump_definitions`, `load_definitions`, `definition_echo`) into `model_dump(mode="json")` and `model_validate(...)`. If you want the dataclass ergonomics specifically, `pydantic.dataclasses.dataclass` is a drop-in for the decorator and gets you the same thing.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/snapshot.py
```diff
@@ -0,0 +1,115 @@
...
+class SnapshotPayload(BaseModel):
+    version: int = SNAPSHOT_VERSION
+    definitions: list[_DefinitionPayload] = Field(default_factory=list)
+
+
+def dump_definitions(definitions: list[CustomFieldDefinition]) -> dict[str, object]:
+    payload = SnapshotPayload(
+        version=SNAPSHOT_VERSION,
+        definitions=[
+            _DefinitionPayload(
```
klcd: Twelve keyword arguments copied one-by-one from `CustomFieldDefinition` into `_DefinitionPayload` — a model whose field names are character-for-character identical. The `list(...)`/`tuple(...)` conversions on `aliases` and `allowed_values` exist only because one side chose tuples and the other lists.

With `CustomFieldDefinition` as a pydantic model this whole function is:

```python
def dump_definitions(definitions: list[CustomFieldDefinition]) -> dict[str, object]:
    return SnapshotPayload(definitions=definitions).model_dump(mode="json")
```

The risk with the current form isn't performance, it's that adding a thirteenth field to `CustomFieldDefinition` requires remembering to touch four places, and forgetting silently drops it from the snapshot rather than failing.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/snapshot.py
```diff
@@ -0,0 +1,115 @@
...
+
+    if parsed.version != SNAPSHOT_VERSION:
+        logger.info(
+            "custom_fields.snapshot.version_mismatch",
+            extra={"found": parsed.version, "expected": SNAPSHOT_VERSION},
+        )
+        return None
+
+    return [
+        CustomFieldDefinition(
```
klcd: The same twelve-field copy in reverse. `load_definitions` is doing the useful work above (version check, `ValidationError` handling) and then spending 18 lines re-assembling objects pydantic could have produced directly:

```python
return parsed.definitions  # once SnapshotPayload holds CustomFieldDefinition
```

Note the version-gating logic on lines 82-87 is worth keeping exactly as-is — that's real behaviour. It's only the reconstruction block that's mechanical.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/responses.py
```diff
@@ -0,0 +1,105 @@
...
+    crm_name: str
+    entity_type: str
+
+
+# Concrete parameterization of the shared model. A plain assignment, not a subclass: pydantic
+# resolves the subscript to a real model class, which is what FastMCP needs for output schemas.
+FieldAmbiguousResponse = AmbiguousResponse[FieldCandidateEcho]
+
+
+def definition_echo(definition: CustomFieldDefinition) -> CustomFieldDefinitionEcho:
```
klcd: Third copy of the same mapping. Even without changing the domain type, pydantic can read attributes off an arbitrary object:

```python
class CustomFieldDefinitionEcho(BaseModel):
    model_config = ConfigDict(from_attributes=True)
    ...

def definition_echo(d: CustomFieldDefinition) -> CustomFieldDefinitionEcho:
    return CustomFieldDefinitionEcho.model_validate(d)
```

That works today against the frozen dataclass and deletes 14 lines. `field_candidate_echo` below is a genuinely different shape (it flattens `candidate.value` and adds `key`/`label`), so that one earns its keep.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/lov.py
```diff
@@ -0,0 +1,246 @@
...
+        if label in self._seen:
+            return
+        self._seen.add(label)
+        self._values.append(AllowedValue(id=entry_id, label=label))
+
+    def build(self) -> tuple[AllowedValue, ...]:
+        return tuple(self._values)
+
+
+def inline_allowed_values(
```
klcd, line 159: This is the largest hand-written JSON parser in the stack: `as_object_dict` / `as_object_list` narrowing at every step, a probe across `_ENTRY_COLLECTION_KEYS`, then an ad-hoc JSON:API unwrap that checks whether `data` is a list or a dict and whether each item has an `attributes` sub-object (`_attributes_or_self`).

Every one of those branches is a shape, and pydantic expresses shapes as types:

```python
class _InlineOption(BaseModel):
    id: str | None = None
    label: str | None = Field(None, validation_alias=AliasChoices(*_LABEL_KEYS))

class _InlineSet(BaseModel):
    entries: list[_InlineOption] = Field(
        default_factory=list, validation_alias=AliasChoices(*_ENTRY_COLLECTION_KEYS)
    )

_INLINE = TypeAdapter(_InlineSet | list[_InlineOption] | None)
```

The JSON:API `data`/`attributes` unwrap is the same thing `backstop_client/json_api.py` already models properly — worth reusing `BackstopApiResource` here rather than re-deriving it from raw dicts.

The practical argument: right now a Backstop instance using a sixth collection key fails silently and returns no picklist options. With a model you get a `ValidationError` naming the payload.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/lov.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def _label_from_option(option: object) -> str | None:
+    text = as_clean_str(option)
+    if text is not None:
+        return text
+    option_dict = as_object_dict(option)
+    if option_dict is None:
+        return None
+    for key in _LABEL_KEYS:
```
klcd: Probing seven keys in priority order is exactly `AliasChoices`, which pydantic resolves left-to-right with the same first-match-wins semantics:

```python
label: str | None = Field(
    default=None,
    validation_alias=AliasChoices("display", "defaultDisplay", "label", "name", "value", "displayName", "display_name"),
)
```

Combined with `StringConstraints(strip_whitespace=True, min_length=1)` that also replaces the `as_clean_str` call inside the loop.

Separately, `_AllowedValueBuilder` (line 63) is a class with `add`, `add_entry`, `build` and a `_seen` set to dedupe by label — that's `{value.label: value for ...}`, since dicts preserve insertion order. Two methods that differ only in whether the label was already extracted suggests the extraction should happen before the builder, not inside it.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/index.py
```diff
@@ -0,0 +1,72 @@
...
+    definitions = index.get(entity, [])
+
+    exact = [d for d in definitions if needle in _searchable_names(d)]
+    if exact:
+        return from_candidates([candidate_for(d) for d in exact], query=query, scope=entity)
+
+    partial = [
+        d
+        for d in definitions
+        if any(needle in name or name in needle for name in _searchable_names(d))
```
klcd: `needle in name or name in needle` is bidirectional substring matching, which has no notion of *how well* something matched — so the partial tier returns an unranked set and any query that's a substring of several field names goes straight to an elicitation prompt. The docstring above acknowledges this ("grade" matching "Investor Grade" and "Grade Review Date") and solves it with tiers, but tiers only separate exact from non-exact; within the partial tier everything is equally good.

Scoring fixes that properly, and stdlib can do it with no new dependency:

```python
matches = difflib.get_close_matches(needle, all_names, n=5, cutoff=0.6)
```

`rapidfuzz.process.extract` is better still (much faster, and `partial_ratio` handles the substring case explicitly) if you're willing to add it. Either way you can then return candidates ordered by score, and `from_candidates` gets a ranked list instead of an arbitrary one.

Also worth noting `name in needle` — the reverse direction — means a one-character field name matches nearly every query.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/index.py
```diff
@@ -0,0 +1,72 @@
...
+    """Resolve a field by human name or alias, exact matches outranking partial ones.
+
+    Two tiers, and the first non-empty one decides: an exact (normalized) name match, then a
+    substring match either way. Tiering matters because a user's short phrase ("grade") is a
+    substring of several instance field names ("Investor Grade", "Grade Review Date") while
+    being an exact match for at most one — without the tiers, the exact hit would be drowned
+    in its own near-misses and every lookup would prompt.
+    """
+    entity = normalize_entity_type(entity_type)
+    needle = normalize_query(query)
```
klcd: Naming: `needle` is haystack-search jargon, and it's the only place in the feature that calls this value anything other than `query`. The shared algebra in `resolution.py` uses `query` throughout — `NotFound.query`, `Ambiguous.query`, `from_candidates(query=...)` — and this function even passes the *original* `query` to those on lines 59/65/72 while matching on `needle`.

So two names for two versions of the same value, distinguished only by normalization. `normalized_query` says which is which. (Same note applies to `haystack` in `backstop_client/errors.py`, which I flagged on #748.)

services/backstop-mcp/src/backstop_mcp/features/custom_fields/service.py
```diff
@@ -0,0 +1,217 @@
...
+
+    @property
+    def is_fresh(self) -> bool:
+        """Whether the in-memory schema came from a fetch recent enough to trust.
+
+        False both when nothing has ever been fetched and when the snapshot has aged past
+        `ttl`. Lets callers decide whether it's worth acquiring a client — which costs a DB
+        read plus a credential decrypt — before asking for a refresh.
+        """
+        return self._fetched_at is not None and datetime.now(UTC) - self._fetched_at < self._ttl
```
klcd: Hand-rolled TTL: a nullable timestamp plus `datetime.now(UTC) - t < ttl`. This exact pattern appears three times in this PR — here, `_within_refresh_floor` at line 171, and the middleware's failure cooldown at `custom_field_glossary.py:127` — each with its own nullable field and its own reset logic.

`cachetools.TTLCache` (or a tiny shared `Expiry` value object wrapping `time.monotonic()`) puts the semantics in one place. Worth doing because the three copies don't currently agree: this one and `_within_refresh_floor` use wall-clock `datetime.now(UTC)`, which jumps under NTP correction — `time.monotonic()` is the right clock for "has an interval elapsed".

Minor naming point: `is_fresh` and `has_definitions` are good predicate names; `_fetched_at` vs `_refresh_attempted_at` needs the four-line comment above it to disambiguate, which suggests `_index_filled_at` vs `_last_fetch_attempt_at`.

services/backstop-mcp/src/backstop_mcp/server/middleware/custom_field_glossary.py
```diff
@@ -0,0 +1,130 @@
...
+            client = await self._client_for_caller()
+            await self._service.ensure_fresh(client)
+        except Exception:
+            self._warm_failed_at = datetime.now(UTC)
+            logger.warning("custom_fields.glossary.warm_failed", exc_info=True)
+
+    def _in_failure_cooldown(self) -> bool:
+        if self._warm_failed_at is None:
+            return False
+        if datetime.now(UTC) - self._warm_failed_at < self.WARM_FAILURE_COOLDOWN:
```
klcd: Third instance of the hand-rolled TTL check (see my note on `service.py:72`) — nullable `_warm_failed_at`, a `timedelta` comparison, and a manual reset on the next line.

Since this one is a circuit-breaker rather than a cache, the cleanest fix is a small shared helper both this and `CustomFieldsService` use:

```python
class Cooldown:
    def __init__(self, period: timedelta) -> None: ...
    def trip(self) -> None: ...
    @property
    def active(self) -> bool: ...
```

Backed by `time.monotonic()`, so it can't be skewed by a clock adjustment.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/fetch.py
```diff
@@ -0,0 +1,116 @@
...
+    lov_index: LovEntryIndex,
+    included: list[dict[str, object]],
+) -> CustomFieldDefinition | None:
+    """Merge one CRM definition with its human override and allowed values.
+
+    Returns None for a definition with no usable name or entity type — neither is recoverable,
+    and an unnameable field can't be resolved by name anyway.
+    """
+    attrs = resource.attributes
+    crm_name = (attrs.name or "").strip()
```
klcd: `(attrs.name or "").strip()` re-strips a value that `CustomFieldDefinitionAttributes` already declared as `_StrippedStr` — pydantic stripped it at validation time. Same on `entity_type` (line 41) and the override aliases (lines 49-54).

If what you actually want is "non-empty after stripping", say that in the model with `StringConstraints(strip_whitespace=True, min_length=1)` and the `or ""` guards go too. Defensive re-stripping is worth removing specifically because it hides whether the model is doing its job — a reader can't tell if these calls are load-bearing.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/values.py
```diff
@@ -0,0 +1,189 @@
...
+    try:
+        return date.fromisoformat(text)
+    except ValueError:
+        pass
+    try:
+        return datetime.fromisoformat(text).date()
+    except ValueError:
+        pass
+    # Last resort for the non-zero-padded forms `date.fromisoformat` rejects (e.g. `2026-9-1`).
+    parts = text.split("-")
```
klcd: A three-tier date parser: `date.fromisoformat`, then `datetime.fromisoformat`, then a manual `split("-")` with `isdigit()` checks to handle non-zero-padded forms like `2026-9-1`.

The third tier is the one to be uneasy about — it accepts `"2026-9-1"` but also silently accepts `"99999-1-1"`, and it's the kind of code that grows a fourth tier the first time an instance writes `09/01/2026`.

Since the value is already flowing through pydantic, this belongs on the model as a `BeforeValidator` on a real `date` field rather than a `str` that gets parsed at sort time. The comment on line 60 explains the `str` choice as avoiding a whole-read failure on one bad value — but `BeforeValidator` returning `None` on unparseable input gives you both: a typed `date | None` field *and* no hard failure.

If you'd rather not hand-write the fallbacks, `dateutil.parser.isoparse` covers tiers one through three on its own.

klcd (commented):
Python-only pass, same lens as #748 / #749 / #716: things hand-rolled where an existing dependency does it, plus naming.

The headline here is that `CustomFieldDefinition` exists **three times**. Once as a frozen dataclass in `types.py`, once as `_DefinitionPayload` in `snapshot.py` for persistence, and once as `CustomFieldDefinitionEcho` in `responses.py` for the wire — with identical field names in all three, and hand-written field-by-field copiers between them (`dump_definitions`, `load_definitions`, `definition_echo`). That's roughly 60 lines of constructor calls whose only job is to move data between shapes that are already the same shape.

Second theme: `lov.py` is a hand-written JSON parser — seven candidate label keys, five candidate collection keys, `as_object_dict`/`as_object_list` narrowing at every step — sitting next to a codebase that models every other wire shape with pydantic.

Third: TTL/expiry arithmetic is re-implemented three times in this PR alone.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/types.py
nicolae536: Done in 7e277105 — `CustomFieldDefinition` / `AllowedValue` are frozen pydantic models; snapshot dump/load uses `model_dump` / `model_validate`.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/snapshot.py
nicolae536: Done in 7e277105 — `dump_definitions` is `SnapshotPayload(...).model_dump(mode="json")`.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/snapshot.py
nicolae536: Done in 7e277105 — `load_definitions` keeps the version/ValidationError gate and returns `parsed.definitions` directly.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/responses.py
nicolae536: Done in 7e277105 — `CustomFieldDefinitionResponse.model_validate(...)` with `from_attributes=True`; renamed Echo → Response.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/lov.py
nicolae536: Done in 7e277105 — inline LOV parsing uses pydantic `_InlineOption` / `_InlineSet` (still tolerates bare strings).

services/backstop-mcp/src/backstop_mcp/features/custom_fields/lov.py
nicolae536: Done in 7e277105 — label probe is `AliasChoices`; builder replaced with insertion-order `dict.setdefault` dedupe.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/index.py
nicolae536: Done in 7e277105 — renamed to `normalized_query`.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/fetch.py
nicolae536: Done in 7e277105 — trust stripped attributes from the model; dropped the defensive re-`.strip()` / `or ""` guards.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/fetch.py
```diff
@@ -0,0 +1,116 @@
...
+    description = (
+        override.description if override is not None and override.description else attrs.description
+    )
+
+    lov_set_ids = resource.related_ids(LOV_SET_RELATIONSHIP)
+    lov_set_id = lov_set_ids[0] if lov_set_ids else None
+
+    return CustomFieldDefinition(
+        definition_id=resource.id,
+        entity_type=entity_type,
```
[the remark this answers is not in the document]
nicolae536: Fixed in 7e277105 — unknown CRM entity types after `normalize_entity_type` are skipped (`return None`) instead of poisoning the snapshot.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/index.py
```diff
@@ -0,0 +1,72 @@
...
+    exact = [d for d in definitions if needle in _searchable_names(d)]
+    if exact:
+        return from_candidates([candidate_for(d) for d in exact], query=query, scope=entity)
+
+    partial = [
+        d
+        for d in definitions
+        if any(needle in name or name in needle for name in _searchable_names(d))
+    ]
+    return from_candidates([candidate_for(d) for d in partial], query=query, scope=entity)
```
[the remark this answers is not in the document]
nicolae536: Fixed in 7e277105 — unknown entity scope returns `NotFound` with the raw scope string; `read_custom_field_value` raises rather than building `/None/...`.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/service.py
```diff
@@ -0,0 +1,217 @@
...
+    async def _load_from_db_unlocked(self) -> None:
+        async with read_session(self._session_factory) as session:
+            snapshot = await load_snapshot(session, self._base_url)
+        if snapshot is None:
+            # Keep whatever is already in memory: an absent or unreadable row is not evidence
+            # that this process's own definitions are wrong.
+            return
+        self._index = build_index(snapshot.definitions)
+        self._fetched_at = snapshot.fetched_at
+        CUSTOM_FIELD_SCHEMA_LOADS.add(1, {"source": "snapshot"})
```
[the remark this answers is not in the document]
nicolae536: Done in 8720a9bf — `apply_overrides` runs on DB load (and after fetch) so config drift applies within TTL.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/index.py
nicolae536: Done in 8720a9bf — partial matches are scored (starts-with > query-in-name > name-in-query); only the best score(s) go to `from_candidates`.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/service.py
nicolae536: Done in 8720a9bf — shared `TimedGate` for freshness / refresh-floor / warm-failure cooldown.

services/backstop-mcp/src/backstop_mcp/server/middleware/custom_field_glossary.py
nicolae536: Done in 8720a9bf — middleware warm failure uses the same `TimedGate`.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/values.py
nicolae536: Done in 8720a9bf — shared `dates.parse_lenient_date` / `LenientDate` (includes non-padded Y-M-D).

services/backstop-mcp/src/backstop_mcp/features/custom_fields/lov.py
```diff
@@ -0,0 +1,284 @@
...
+
+class _InlineSet(BaseModel):
+    """An inlined LOV set object carrying options under one of several collection keys."""
+
+    model_config: ClassVar[ConfigDict] = ConfigDict(extra="ignore")
+
+    entries: list[object] = Field(
+        default_factory=list,
+        validation_alias=AliasChoices(*_ENTRY_COLLECTION_KEYS),
+    )
```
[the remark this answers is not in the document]
nicolae536: Done in 8720a9bf — entry collections pick the first non-empty list among known keys (no AliasChoices short-circuit).

services/backstop-mcp/src/backstop_mcp/features/custom_fields/lov.py
```diff
@@ -0,0 +1,284 @@
...
+        parsed = _InlineOption.model_validate(payload)
+    except ValidationError:
+        return None
+
+    if parsed.label is None:
+        if parsed.id is None:
+            return None
+        return AllowedValue(id=str(parsed.id), label=str(parsed.id))
+    option_id = str(parsed.id) if parsed.id is not None else None
+    return AllowedValue(id=option_id, label=parsed.label)
```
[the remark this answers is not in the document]
nicolae536: Done in 8720a9bf — JSON:API option unwrap keeps the envelope `id`.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/warmup.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    if credential is None:
+        logger.info("custom_fields.warmup.skipped", extra={"reason": "no_service_account"})
+        return
+
+    try:
+        await service.ensure_fresh(
+            clients.for_credential(credential),
+            subject=credential.username,
+        )
```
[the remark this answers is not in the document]
nicolae536: Done in 2373be95 — under subject-keyed snapshots a service-account fill cannot seed end-user catalogs. Warmup is now a `/system-info` connectivity probe; each caller's first authenticated tools/list or list_custom_fields fills their own cache.

services/backstop-mcp/tests/server/middleware/test_custom_field_glossary.py
```diff
@@ -0,0 +1,357 @@
...
+            EntityType.ORGANIZATIONS,
+        )
+
+        person_meta = getattr(get_person, "__fastmcp__", None)
+        assert isinstance(person_meta, ToolMeta)
+        assert parse_glossary_entities(_as_object_dict(person_meta.meta)) == (EntityType.PEOPLE,)
+
+        list_meta = getattr(list_custom_fields, "__fastmcp__", None)
+        assert isinstance(list_meta, ToolMeta)
+        assert parse_glossary_entities(_as_object_dict(list_meta.meta)) == ()
```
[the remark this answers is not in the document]
nicolae536: Not an issue as written — the test guards with `pytest.importorskip("backstop_mcp.server.tools.get_organization")`, so on this PR's branch (where the tool modules genuinely don't exist yet) it skips rather than failing collection. The modules and a populated `TOOLS` land in #751, where the assertion actually runs.

services/backstop-mcp/src/backstop_mcp/app.py
```diff
@@ -122,7 +151,9 @@ async def lifespan(_server: FastMCP) -> AsyncGenerator[None, None]:
     @mcp.custom_route("/ready", methods=["GET"])
     async def ready(_request: Request) -> JSONResponse:
         """Postgres readiness — stock `setup_ops` `/probe` is process-up only."""
-        return await _ready_response(engine)
```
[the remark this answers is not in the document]
nicolae536: No longer applies — warmup was reworked in 4aa87b96. `warm_custom_field_schema` is now a `/system-info` connectivity probe that touches neither `CustomFieldsService` nor the database, so there's no `ensure_fresh`-against-a-disposed-engine path left.

The ordering you described is unchanged (the lifespan `finally` still runs before `warmup_lifespan` exits), so a probe in flight at shutdown can hit a closed httpx client — but the warmup catches and logs that itself, so the worst case is a noisy shutdown log rather than a failure.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/service.py
```diff
@@ -0,0 +1,217 @@
...
+                if not self.has_definitions:
+                    raise
+                logger.warning(
+                    "custom_fields.schema.refresh_failed_serving_stale",
+                    extra={
+                        "fetched_at": (self._fetched_at.isoformat() if self._fetched_at else None),
+                    },
+                    exc_info=True,
+                )
+                CUSTOM_FIELD_SCHEMA_LOADS.add(1, {"source": "stale"})
```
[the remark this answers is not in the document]
nicolae536: Fixed in 6f1f9525 — `ensure_fresh` now consults `entry.refresh_floor.within()` before attempting the fetch, so a failed refresh no longer leaves every subsequent call re-hitting Backstop. `refresh()` keeps its own floor check; `_refresh_unlocked` still stamps the floor for both.

services/backstop-mcp/src/backstop_mcp/features/custom_fields/overrides.py
```diff
@@ -0,0 +1,48 @@
...
+    return entity_type, crm_name
+
+
+def index_overrides(overrides: dict[str, FieldOverride]) -> OverrideIndex:
+    """Map (normalized_entity_type, crm_name) → override."""
+    indexed: OverrideIndex = {}
+    for key, override in overrides.items():
+        entity_type, crm_name = parse_override_key(key)
+        indexed[(normalize_entity_type(entity_type), crm_name)] = override
+    return indexed
```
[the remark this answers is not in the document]
nicolae536: Correcting myself — this one is already handled, and my "still open" reply was wrong.

I checked `_field_overrides` and stopped there. The validation does run at startup, one level down: `create_app` → `create_custom_fields_service` → `CustomFieldsService.__init__`, which calls `index_overrides(overrides)` (service.py). A malformed `entityType:crmName` key or an unknown entity type raises there, before any fetch — so the service never reaches `fetch_custom_field_definitions` with a bad key.

No change needed.

services/backstop-mcp/src/backstop_mcp/app.py
```diff
@@ -165,11 +196,46 @@ def retry_settings(config: BackstopConfig) -> RetrySettings:
...
+    `features/custom_fields` stay free of any `config` import.
+    """
+    return {
+        key: FieldOverride(
+            display_name=override.display_name,
+            aliases=tuple(override.aliases),
+            description=override.description,
+        )
+        for key, override in configured.items()
+    }
```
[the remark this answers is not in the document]
nicolae536: Correcting myself — same as the `overrides.py` instance of this: already handled, and my "still open" reply was wrong.

`_field_overrides` indeed only maps the values across, but the index is built in `CustomFieldsService.__init__` (`index_overrides(overrides)`), which `create_app` calls during startup. So a malformed key fails the boot rather than the refresh.

No change needed.

services/backstop-mcp/src/backstop_mcp/server/middleware/custom_field_glossary.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        self._warm_failure.clear_if_expired()
+        if self._warm_failure.within():
+            return
+        try:
+            client = await self._client_for_caller()
+            await self._service.ensure_fresh(client, subject=subject)
+        except Exception:
+            self._warm_failure.mark()
+            logger.warning("custom_fields.glossary.warm_failed", exc_info=True)
```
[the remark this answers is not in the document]
nicolae536: Fixed in 6f1f9525 — `_warm_failures` is now a `dict[str, TimedGate]` keyed by subject, checked and marked per caller, and cleared on a successful warm so a spent gate doesn't linger. One caller's expired credential no longer blocks anyone else's warm.

services/backstop-mcp/src/backstop_mcp/server/tools/utils/get_activity_history_utils.py
```diff
@@ -0,0 +1,310 @@
...
+    """Minimal attributes this tool needs from the party fetch: a display name plus provenance.
+
+    `extra="ignore"`, not `"allow"` — unlike `get_person`/`get_organization`, this tool never
+    surfaces the raw attribute dump, only `name` (for the resolve echo) and provenance (for
+    `as_of`).
+    """
+
+    model_config: ClassVar[ConfigDict] = ConfigDict(extra="ignore", populate_by_name=True)
+
+    name: str | None = None
```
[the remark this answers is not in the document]
nicolae536: This finding was leftover from the mashed #750 tip. Addressed on #794: activity-history `PartyAttributes` now keeps `firstName`/`lastName` so a `type="next"` page can rebuild the display name from the party GET.

vabatta (approved):

PR #751, opened 2026-08-06
feat(backstop-mcp): add data hygiene and MCP tool surface

nicolae536:
## Summary

- Data hygiene: departed-contact detection via employment relationship index + `as_of` provenance on reads ([UN-23678](https://unique-ch.atlassian.net/browse/UN-23678))
- Registers the MCP tool surface that composes the stacked libraries:
  - `get_system_info`
  - `get_organization` (party resolve → fetch + glossary meta + `as_of`)
  - `get_person` (party resolve → fetch + departed flag + `as_of`)
  - `list_custom_fields`
- Final tip of this stack matches the pre-split backup tree (`backstop-mcp/feat/UN-22647--scafolding-backup-20260806`)

## Stack

1. Scaffolding (#716)
2. Client + Auth (#748)
3. Party resolver (#749)
4. Custom fields (#750) ← base
5. **This PR (#751)** — data hygiene + all tools

## Review focus

- Employment index / former-employment markers vs ids
- `as_of` provenance (no staleness verdict)
- How tools compose party + custom fields + hygiene

## Test plan

- [ ] `cd services/backstop-mcp && uv sync && uv run ruff check . && uv run basedpyright && uv run pytest`
- [ ] `get_organization` with fuzzy name and trusted `party_id`
- [ ] Ambiguous name elicits a choice (or structured ambiguous payload)
- [ ] `list_custom_fields` / glossary on `tools/list`
- [ ] `get_person` reports departed contacts when former employment is present

[UN-23678]: https://unique-ch.atlassian.net/browse/UN-23678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/services/backstop-mcp/README.md
+++ b/services/backstop-mcp/README.md
@@ -9,12 +9,8 @@ a Backstop username + personal API token. That credential is verified against Ba
 never as a shared service account. Failed logins are rate-limited per username
 (`AUTH_LOGIN_MAX_ATTEMPTS`) so the form can't be used to test credentials against Backstop.
 
-This PR adds CRM custom-field schema discovery and caching (`features/custom_fields`) on top of
-the Backstop client, the OAuth/login bridge, and the `party_resolver` name-to-record lookup
-library. The schema is fetched lazily (or warmed at startup with an optional service account),
-cached in Postgres with a TTL, and enriched with human-facing glossaries per entity type. No MCP
-tools are exposed yet — `list_custom_fields`, `get_organization`, `get_person`, and
-`get_system_info` land in a later PR.
+Tools resolve records by name rather than by ID: "Capstone" or "Investor Status" is looked up
+against the live instance, and an ambiguous match asks the user to pick one.
 
 ## Layout
 
@@ -25,21 +21,19 @@ src/backstop_mcp/
   logging.py metrics.py  cross-cutting, used by both sides below
   features/              what the connector does
     resolution.py          the shared "which record is that?" algebra + ambiguity policy
-    entity_types.py        canonical Backstop entity-type vocabulary
     auth/                  Backstop credential bridging: login form, encryption, token rotation
     custom_fields/         CRM custom-field schema discovery, caching and resolution
     party_resolver/        name / email / trusted-ID lookup for organizations, people, contacts
-  server/                how it's exposed over MCP — no tools registered yet
-    runtime.py             the process-wide service holder tools will reach through
-    tools/                 empty registry until the first tool lands
-    middleware/            the custom-field glossary middleware (tools/list enrichment)
+  server/                how it's exposed over MCP
+    runtime.py             the process-wide service holder tools reach through
+    tools/                 tool functions + the single registry declaring them
+    middleware/            FastMCP and ASGI middleware
   backstop_client/       HTTP transport: auth headers, concurrency gate, retries, pagination
   db/                    SQLAlchemy models, engine, alembic migrations
 ```
 
 The layering rule is that **nothing under `features/` may import from `server/`** — the server
-wires features together, never the reverse. `tests/test_layering.py` enforces it, and grows as
-each package arrives.
+wires features together, never the reverse. `tests/test_layering.py` enforces it.
 
 ## Run locally
 
@@ -54,8 +48,7 @@ uv run backstop-mcp
 - MCP endpoint: `http://localhost:9010/mcp` (HTTP transport)
 - Health: `GET /health` — liveness via `unique_mcp.monitoring.setup_ops`
 - Probe: `GET /probe` — process-up (setup_ops)
-- Ready: `GET /ready` — 503 when Postgres is unreachable; also reports whether the custom-field
-  schema has loaded (never gates readiness — it fills lazily by design)
+- Ready: `GET /ready` — 503 when Postgres is unreachable
 - Metrics: `GET /metrics` — Prometheus (setup_ops)
 
 Generate an encryption key with:
--- a/services/backstop-mcp/README.md
+++ b/services/backstop-mcp/README.md
@@ -21,6 +21,7 @@ src/backstop_mcp/
   logging.py metrics.py  cross-cutting, used by both sides below
   features/              what the connector does
     resolution.py          the shared "which record is that?" algebra + ambiguity policy
+    entity_types.py        canonical Backstop entity-type vocabulary
     auth/                  Backstop credential bridging: login form, encryption, token rotation
     custom_fields/         CRM custom-field schema discovery, caching and resolution
     party_resolver/        name / email / trusted-ID lookup for organizations, people, contacts
@@ -49,7 +50,9 @@ uv run backstop-mcp
 - MCP endpoint: `http://localhost:9010/mcp` (HTTP transport)
 - Health: `GET /health` — liveness via `unique_mcp.monitoring.setup_ops`
 - Probe: `GET /probe` — process-up (setup_ops)
-- Ready: `GET /ready` — 503 when Postgres is unreachable or custom-field schema is not loaded
+- Ready: `GET /ready` — 503 when Postgres is unreachable. Also reports whether the
+  custom-field schema has loaded, which never gates readiness: it fills lazily per caller,
+  and tools degrade to `list_custom_fields` without it
 - Metrics: `GET /metrics` — Prometheus (setup_ops)
 
 Generate an encryption key with:
--- a/services/backstop-mcp/README.md
+++ b/services/backstop-mcp/README.md
@@ -24,6 +24,7 @@ src/backstop_mcp/
     auth/                  Backstop credential bridging: login form, encryption, token rotation
     custom_fields/         CRM custom-field schema discovery, caching and resolution
     party_resolver/        name / email / trusted-ID lookup for organizations, people, contacts
+    data_hygiene/          employment edges, departed contacts, as-of provenance
   server/                how it's exposed over MCP
     runtime.py             the process-wide service holder tools reach through
     tools/                 tool functions + the single registry declaring them
@@ -48,7 +49,7 @@ uv run backstop-mcp
 - MCP endpoint: `http://localhost:9010/mcp` (HTTP transport)
 - Health: `GET /health` — liveness via `unique_mcp.monitoring.setup_ops`
 - Probe: `GET /probe` — process-up (setup_ops)
-- Ready: `GET /ready` — 503 when Postgres is unreachable
+- Ready: `GET /ready` — 503 when Postgres is unreachable or custom-field schema is not loaded
 - Metrics: `GET /metrics` — Prometheus (setup_ops)
 
 Generate an encryption key with:
--- a/services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
+++ b/services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
@@ -5,20 +5,12 @@ any of it directly — they go through `service.EmploymentIndexFactory`, which o
 vocabulary. List/org-contact tools should use that verdict to exclude departed people
 from "who do we contact at X" answers unless the user asked for historical contacts; a by-id
 person fetch returns the person with the flag set rather than hiding the record.
-
-`_employment_edges` is the shared building block for reading employment off *either* side of a
-relationship — a person's own GET or an organization's own GET — with `person_side` as the only
-difference between the two directions. It is one step below `EmploymentIndex` (a later addition):
-it normalises the raw payload into edges, but does not yet reduce several edges for the same pair
-down to one winner.
 """
 
 from collections.abc import Sequence
 from dataclasses import dataclass
-from datetime import date, datetime
+from datetime import date
 from typing import TypeGuard
-
-from pydantic import ValidationError
 
 from backstop_mcp.backstop_client import BackstopApiResource
 from backstop_mcp.coerce import as_clean_str
@@ -37,6 +29,9 @@ from backstop_mcp.features.data_hygiene.types import (
     RelationshipTypeAttributes,
 )
 from backstop_mcp.features.entity_types import normalize_entity_type
+
+type RelationshipResource = BackstopApiResource[EntityRelationshipAttributes]
+type RelationshipTypeResource = BackstopApiResource[RelationshipTypeAttributes]
 
 
 @dataclass(frozen=True)
@@ -112,8 +107,7 @@ def _rank(edge: EmploymentEdge) -> tuple[bool, date, bool]:
 
 
 def _to_record(edge: EmploymentEdge) -> EmploymentRecord:
-    departure = edge.evidence if edge.status is EmploymentStatus.FORMER else None
-    return EmploymentRecord(status=edge.status, departure=departure)
+    return EmploymentRecord(status=edge.status, departure=edge.departure)
 
 
 def classify_employment(
@@ -129,8 +123,9 @@ def classify_employment(
       employee of` contains `employee`); an employment-first test would call it `CURRENT` and
       report someone who has left as a live contact.
     * `IRRELEVANT` — employment vocabulary is configured, the type has a name to judge, and it
-      matches nothing.
-    * `CURRENT` — otherwise.
+      matches nothing — or the type id is present but its name did not side-load, so we cannot
+      treat it as employment evidence (a portal-style edge must not clear a real departure).
+    * `CURRENT` — otherwise (including truly untyped relationships, so `endDate` still applies).
     """
     if rules.former.matches(type_id=type_id, type_name=type_name):
         return EmploymentStatus.FORMER
@@ -140,42 +135,21 @@ def classify_employment(
         return EmploymentStatus.CURRENT
     if rules.employment.matches(type_id=type_id, type_name=type_name):
         return EmploymentStatus.CURRENT
-    if type_name is None:
-        # Nothing to judge: either the record carries no type, or its type resource did not
-        # side-load. Both leave the person→org gate below as the only evidence, so read it as
-        # current — that keeps `endDate` in play and never invents a departure out of a record we
-        # could not classify.
+    if type_name is None and type_id is None:
+        # Truly untyped: leave the person→org gate as the only evidence and read it as current
+        # so `endDate` stays in play.
         return EmploymentStatus.CURRENT
     return EmploymentStatus.IRRELEVANT
 
 
-def _relationship_type_names(*, resources: list[dict[str, object]]) -> dict[str, str]:
-    """`id → name` for the side-loaded relationship types. Unnamed and malformed ones are dropped.
-
-    A type we cannot read leaves its relationships with no name to match, which
-    `classify_employment` treats as no type signal — never as a positive finding.
-    """
+def _relationship_type_names(*, resources: Sequence[RelationshipTypeResource]) -> dict[str, str]:
+    """`id → name` for the side-loaded relationship types. Unnamed ones are dropped."""
     names: dict[str, str] = {}
-    for raw in resources:
-        try:
-            resource = BackstopApiResource[RelationshipTypeAttributes].model_validate(raw)
-        except ValidationError:
-            continue
+    for resource in resources:
         name = as_clean_str(resource.attributes.name)
         if name is not None:
             names[resource.id] = name
     return names
-
-
-def _safe_parse_relationship(
-    *, raw: dict[str, object]
-) -> tuple[EntityRelationshipAttributes, str | None] | None:
-    try:
-        resource = BackstopApiResource[EntityRelationshipAttributes].model_validate(raw)
-    except ValidationError:
-        return None
-    type_ids = resource.related_ids(EntityRelationshipRef.TYPE)
-    return resource.attributes, type_ids[0] if type_ids else None
 
 
 def _side_type(*, side: EntityRefAttributes) -> str | None:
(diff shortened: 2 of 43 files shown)


klcd (commented):
Python-only pass, same lens as the rest of the stack: hand-rolled code where a dependency already solves it, plus naming.

Two things stand out beyond the usual pydantic/dataclass duplication.

First, there is dead configuration in the employment pipeline. `person_side` is threaded through three functions, documented at length in two docstrings, and then discarded with `_ = person_side` on the first line of the function that receives it — which makes `build_person_employment_index` and `build_organization_employment_index` byte-identical apart from the literal they pass. Two public functions, one behaviour.

Second, `EmploymentEdge.evidence` is typed `DepartedEmployment` even on `CURRENT` edges, where the code fills it with placeholder values it explicitly documents as "not a claim about anything". A type whose name asserts something false for a whole branch of its inputs is worth fixing before this grows more callers.

Also: dates are carried as `str` on pydantic models and parsed later by hand, and the tool surface leaks `dict[str, object]` where validated models already exist.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
```diff
@@ -0,0 +1,412 @@
...
+    each landed on, so both `person_id` and `organization_id` are pulled from whichever side
+    matched — `person_side` only needs to be threaded through for callers building an index (a
+    later addition) that must record which id belongs to which entity type.
+
+    `IRRELEVANT` edges are dropped — they neither vouch for the person nor speak against them, so
+    they carry no employment signal for `EmploymentIndex` to fold. An edge with no usable date at
+    all (`effective_date=None`) is kept: it sorts last downstream rather than being dropped
+    outright, so it still wins when it is the only edge for its pair.
+    """
+    _ = person_side
```
klcd: `_ = person_side` discards the parameter on the first line of the function. The docstring above spends five lines explaining that structural matching is direction-agnostic and that `person_side` "only needs to be threaded through for callers building an index (a later addition)" — but no such caller exists in this PR, and `_employer_side`/`_person_id` both derive the sides from entity *type* rather than from position.

The knock-on is `build_person_employment_index` and `build_organization_employment_index` (lines 369 and 393), which are identical except for `person_side=True` vs `False`. Its own docstring says so: "the only real difference between the two builders is which literal value they pass."

Delete the parameter and collapse the two builders into `build_employment_index(...)`. `EmploymentIndexFactory` can keep its two named entry points if the call sites read better that way, but they should not be two implementations. If the flag is genuinely needed by upcoming work, add it in the PR that uses it — right now it's an unused parameter that ruff's ARG rules would flag if they were enabled, which is presumably why the `_ =` is there.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
```diff
@@ -0,0 +1,412 @@
...
+                end_date=ended.isoformat(),
+                relationship_type_id=type_id,
+                relationship_type_name=type_name,
+            )
+        else:
+            effective_date = started if started is not None else created
+            # `evidence` is only ever read by the resolver (a later addition) for a `FORMER`
+            # edge; a plain `CURRENT` edge has no departure to describe, so `signal`/`end_date`
+            # here are a placeholder to satisfy the required field, not a claim about anything.
+            evidence = DepartedEmployment(
```
klcd: This constructs a `DepartedEmployment` for an edge that is explicitly **not** a departure — `signal=DepartureSignal.END_DATE` and `end_date=None` on a `CURRENT` edge. The comment above is candid that these are "a placeholder to satisfy the required field, not a claim about anything", which is the tell: the type is being bent to fit a field that shouldn't require it.

The union you actually have is "this edge is current" or "this edge ended, and here's how we know". Model it:

```python
class EmploymentEdge(BaseModel):
    ...
    departure: DepartedEmployment | None = None  # set iff status is FORMER
```

Organization identity (`organization_id`, `organization_type`, `relationship_type_id/name`) is on the edge in all three branches anyway, so it belongs on `EmploymentEdge` rather than being duplicated into the evidence object each time — that also removes the six repeated keyword arguments across the three branches here.

As it stands, any future reader who sees `edge.evidence.signal` has no way to know it's meaningless unless `edge.status is FORMER`.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/types.py
```diff
@@ -0,0 +1,192 @@
...
+    existing `DepartedEmployment` payload so tool responses keep their current shape even for a
+    `CURRENT` edge, where the payload describes the relationship rather than an actual departure.
+    """
+
+    person_id: str
+    organization_id: str
+    organization_type: str
+    status: EmploymentStatus
+    effective_date: date | None
+    evidence: DepartedEmployment
```
klcd: Naming follow-on from `employment.py:320`: the field is called `evidence` and typed `DepartedEmployment`, so both the name and the type assert a departure — but on `CURRENT` edges it holds placeholder data.

If you take the `departure: DepartedEmployment | None` shape, the name problem solves itself. If you'd rather keep one field, `relationship_context` describes what it actually holds in all three cases.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/types.py
```diff
@@ -0,0 +1,192 @@
...
+    resource_id: str | None = Field(default=None, alias="resourceId")
+    resource_type: str | None = Field(default=None, alias="resourceType")
+
+
+class EntityRelationshipAttributes(BaseModel):
+    """Sparse attributes used to decide whether a person→org employment has ended."""
+
+    model_config: ClassVar[ConfigDict] = ConfigDict(extra="ignore")
+
+    end_date: str | None = Field(default=None, alias="endDate")
```
klcd: `end_date` (and `start_date`, `created_timestamp`) are typed `str` on a pydantic model, which is why `employment.py:221` needs a hand-written `_safe_parse_date`. Pydantic coerces ISO strings to `date`/`datetime` natively, and a `BeforeValidator` covers Backstop's odd spellings:

```python
def _lenient_date(v: object) -> object:
    return v[:10] if isinstance(v, str) and len(v) > 10 else v

end_date: Annotated[date | None, BeforeValidator(_lenient_date)] = Field(None, alias="endDate")
```

That deletes `_safe_parse_date` and, more usefully, removes the `.isoformat()` round-trip at `employment.py:298` where a parsed date is turned back into a string to store on `DepartedEmployment`.

Separate point on the same line: `alias="endDate"` is spelled out per field throughout this model. `model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True)` derives all of them — same suggestion I left on `backstop_client/pagination.py` in #748, and it's worth a shared base model for JSON:API attribute types across the stack.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
```diff
@@ -0,0 +1,412 @@
...
+        return None
+    return _Employer(organization_id=organization_id, organization_type=organization_type)
+
+
+def _is_organization(side_type: str | None) -> TypeGuard[str]:
+    """`TypeGuard` rather than a bare `in`, so the matched side's type reads as the `str` it is."""
+    return side_type in ORG_SIDE_TYPES
+
+
+def _safe_parse_date(*, raw: str | None) -> date | None:
```
klcd: This exists only because the model stores dates as `str` (see my note on `types.py:54`). Once the fields are typed `date | None`, the whole function and its three call sites on lines 288-290 go away.

Worth flagging the naming too: `_safe_parse_date` says nothing about *what* is safe. It means "returns None instead of raising" — `_parse_date_or_none` is the convention that reads unambiguously. `raw` as a keyword-only parameter name is also vague for something that's specifically a date string.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/entity_relationships.py
```diff
@@ -0,0 +1,32 @@
...
+from backstop_mcp.backstop_client import (
+    BackstopApiResourceDocument,
+    follow_included,
+    included_by_type,
+)
+from backstop_mcp.features.data_hygiene.types import EntityRelationshipRef
+
+
+class EntityRelationships(TypedDict):
+    relationships: list[dict[str, object]]
```
klcd: `list[dict[str, object]]` carries the side-loaded relationships all the way through the pipeline, and every consumer re-validates them individually — `_safe_parse_relationship` inside the loop at `employment.py:272`, and the same for relationship types in `_relationship_type_names`.

Validate once, at the boundary where the data arrives:

```python
relationships: list[BackstopApiResource[EntityRelationshipAttributes]]
```

`backstop_client.json_api` already provides `BackstopApiResource[AttrT]` and `follow_included`/`included_by_type` for exactly this. Then `_employment_edges` takes typed resources, the per-item try/except disappears, and a malformed relationship surfaces as one `ValidationError` naming the field rather than being silently `continue`d past.

services/backstop-mcp/src/backstop_mcp/server/tools/get_person.py
```diff
@@ -0,0 +1,152 @@
...
+
+    Always returns the person when resolved. When employment has ended anywhere, `departed` is
+    true and `departures` carries one entry per organization the person has left, each already
+    identifying its own `organization_id` — relay every entry to the user and do not present the
+    person as a current contact at any of those organizations unless they asked for historical
+    contacts.
+    """
+
+    status: Literal["resolved"] = "resolved"
+    person: dict[str, object]
```
klcd: `person: dict[str, object]` on the response model throws away the validation that just happened — `PersonAttributes` parsed this data a few lines earlier. FastMCP builds its structured-output schema from these models, so the tool advertises an opaque object to the client instead of the field list it actually returns.

`person: PersonAttributes` (with `extra="allow"` on that model if instance-specific fields need to pass through) keeps the schema meaningful. Same shape of issue in `get_organization.py`.

services/backstop-mcp/src/backstop_mcp/server/tools/get_organization.py
```diff
@@ -0,0 +1,123 @@
...
+        party_id=party_id,
+        search=search,
+    )
+    if not isinstance(result, Resolved):
+        return tool_result(unresolved_party_response(result))
+
+    party = result.value
+    path = f"/organizations/{quote(party.id, safe='')}"
+    document = await client.get(path, schema=BackstopApiResourceDocument[OrganizationAttributes])
+    attributes = document.data.attributes.model_dump(exclude_none=True)
```
klcd: Validate → serialize → re-parse, in three consecutive steps: `document.data.attributes` is a validated `OrganizationAttributes`, `.model_dump(exclude_none=True)` flattens it back to a dict, and then `extract_as_of(attributes)` re-reads `"modifiedTimestamp"` / `"modifiedBy"` out of that dict by string key.

The string keys are the part that bites: nothing connects them to the model's field names, so renaming a field on `OrganizationAttributes` leaves `extract_as_of` silently returning `None`.

A small `ProvenanceFields` mixin (`modified_timestamp`, `modified_by`, with the camelCase aliases) that both attribute models inherit would let `extract_as_of` take the model and read attributes — checked by the type checker rather than by hope.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/types.py
```diff
@@ -0,0 +1,192 @@
...
+    destination_entity: EntityRefAttributes | None = Field(default=None, alias="destinationEntity")
+
+
+class RelationshipTypeAttributes(BaseModel):
+    model_config: ClassVar[ConfigDict] = ConfigDict(extra="ignore")
+
+    name: str | None = None
+
+
+@dataclass(frozen=True)
```
klcd: `AsOf` is a frozen dataclass; `AsOfEcho` in `responses.py` is a pydantic model with the same fields; `as_of_echo()` at `responses.py:35` copies one into the other. Same three-part pattern I flagged on #750 for `CustomFieldDefinition` and on #749 for the resolution algebra — it's now the most repeated structural issue in the stack.

One pydantic model serves both roles here. Failing that, `AsOfEcho.model_validate(value, from_attributes=True)` replaces the copier today without touching `AsOf`.

`as_of_echo` and `departed_echo` (line 44 of `responses.py`) are also both one-line field copies whose parameter is named `value` — worth naming for what it is if they survive.

services/backstop-mcp/src/backstop_mcp/server/tools/results.py
```diff
@@ -0,0 +1,22 @@
...
+
+import json
+
+from mcp.types import CallToolResult, TextContent
+from pydantic import BaseModel
+
+
+def tool_result(payload: BaseModel | dict[str, object]) -> CallToolResult:
+    """Serialize a domain payload as a single JSON text content block."""
+    text = payload.model_dump_json() if isinstance(payload, BaseModel) else json.dumps(payload)
```
klcd: Accepting `dict[str, object]` alongside `BaseModel` means a tool can return unvalidated, unschematized JSON — and `get_system_info` does exactly that. Since the module docstring states the intent as "tools keep domain Pydantic models internally", the signature is undercutting its own contract.

Narrowing to `payload: BaseModel` makes the rule enforceable at the type level and the `json.dumps` branch disappears. `get_system_info` needs a small response model, which it should have anyway so FastMCP can publish its output schema.

One behavioural note if you do: `model_dump_json()` defaults to `by_alias=False`, so any model using camelCase aliases will serialize with Python field names. Worth deciding that explicitly here rather than per-tool.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
nicolae536: Done in f74bfa3f — dropped `person_side` and collapsed the two builders into `build_employment_index`; the factory still exposes the named entry points as thin wrappers.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
nicolae536: Done in f74bfa3f — `EmploymentEdge.departure` is now `DepartedEmployment | None`; CURRENT edges no longer carry a placeholder departure.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/types.py
nicolae536: Done in f74bfa3f — renamed/retyped with the optional `departure` field so CURRENT edges don't pretend to be departed.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/types.py
nicolae536: Done in f74bfa3f — relationship date fields are `date | None` via a lenient `BeforeValidator` (`_lenient_date`).

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
nicolae536: Done in f74bfa3f — `_safe_parse_date` is gone; classification reads typed `date` fields directly.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/entity_relationships.py
nicolae536: Done in f74bfa3f — side-loaded relationships parse once to typed `BackstopApiResource[...]` (bad includes soft-skipped with a warning).

services/backstop-mcp/src/backstop_mcp/server/tools/get_person.py
nicolae536: Done in f74bfa3f — `PersonResolvedResponse.person` is typed `PersonAttributes` (with `ProvenanceFields`).

services/backstop-mcp/src/backstop_mcp/server/tools/get_organization.py
nicolae536: Done in f74bfa3f — organization attributes stay as the validated model on the response; no dump/re-parse round trip for the payload.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/types.py
nicolae536: Done in f74bfa3f — `AsOf` is a pydantic model; `as_of_response` / `DepartedContactResponse` replace the Echo copy layer.

services/backstop-mcp/src/backstop_mcp/server/tools/results.py
nicolae536: Done in f74bfa3f — `tool_result` accepts `BaseModel` only; `get_system_info` returns `SystemInfoResponse`.

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/employment.py
```diff
@@ -0,0 +1,412 @@
...
+        # `EmploymentRules` for what that costs.
+        return EmploymentStatus.CURRENT
+    if rules.employment.matches(type_id=type_id, type_name=type_name):
+        return EmploymentStatus.CURRENT
+    if type_name is None:
+        # Nothing to judge: either the record carries no type, or its type resource did not
+        # side-load. Both leave the person→org gate below as the only evidence, so read it as
+        # current — that keeps `endDate` in play and never invents a departure out of a record we
+        # could not classify.
+        return EmploymentStatus.CURRENT
```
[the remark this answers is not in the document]
nicolae536: Done in f74bfa3f — when a type id is present but the name is missing (and vocabulary is configured), classify returns `IRRELEVANT` instead of treating the edge as CURRENT, so an untyped/unsideloaded edge can't clear a real FORMER winner.

services/backstop-mcp/src/backstop_mcp/server/tools/list_custom_fields.py
```diff
@@ -0,0 +1,70 @@
...
+    accounts. Pass refresh=true to re-fetch definitions from Backstop into the cache.
+    """
+    client = await get_backstop_client()
+    service = get_custom_fields_service()
+    if refresh:
+        await service.refresh(client)
+    else:
+        await service.ensure_fresh(client)
+
+    definitions = service.definitions_for(entity_type.value)
```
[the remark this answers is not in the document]
nicolae536: Done in 8720a9bf (schema service) + c0bf690c (tool wiring) — snapshots and in-memory indexes are keyed by `(base_url, subject)`; warmup uses the service-account username; `list_custom_fields` / glossary use `current_subject()`.

services/backstop-mcp/README.md
```diff
@@ -25,21 +21,19 @@ src/backstop_mcp/
   logging.py metrics.py coerce.py    cross-cutting, used by both sides below
   features/              what the connector does
     resolution.py          the shared "which record is that?" algebra + ambiguity policy
-    entity_types.py        canonical Backstop entity-type vocabulary
     auth/                  Backstop credential bridging: login form, encryption, token rotation
     custom_fields/         CRM custom-field schema discovery, caching and resolution
     party_resolver/        name / email / trusted-ID lookup for organizations, people, contacts
-  server/                how it's exposed over MCP — no tools registered yet
-    runtime.py             the process-wide service holder tools will reach through
```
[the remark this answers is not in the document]
nicolae536: Fixed in f639b0d2 — `entity_types.py` is back in the layout list. (`data_hygiene/` was already there from the earlier pass.)

services/backstop-mcp/src/backstop_mcp/features/data_hygiene/types.py
```diff
@@ -0,0 +1,192 @@
...
+from typing import ClassVar
+
+from pydantic import BaseModel, ConfigDict, Field
+
+# Which resource types can hold an employment relationship, as the canonical plurals that
+# `entity_types.normalize_entity_type` maps to — `departed.py` compares through it. Backstop
+# emits the API path segment (`people`, `organizations`) and an admin cannot rename those, so
+# this is product schema rather than tenant vocabulary and has no business being configurable:
+# a deployment cannot know the strings, and a typo would silently disable detection.
+PERSON_SIDE_TYPES: frozenset[str] = frozenset({"people", "contacts", "employees"})
```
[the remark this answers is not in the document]
nicolae536: The cosmetic leftover is fixed too, in f639b0d2 — `test_layering.py` no longer names `data_hygiene.departed`; the fixtures use a module that actually exists, so the comment describing them is true again.

services/backstop-mcp/README.md
```diff
@@ -54,8 +48,7 @@ uv run backstop-mcp
 - MCP endpoint: `http://localhost:9010/mcp` (HTTP transport)
 - Health: `GET /health` — liveness via `unique_mcp.monitoring.setup_ops`
 - Probe: `GET /probe` — process-up (setup_ops)
-- Ready: `GET /ready` — 503 when Postgres is unreachable; also reports whether the custom-field
-  schema has loaded (never gates readiness — it fills lazily by design)
+- Ready: `GET /ready` — 503 when Postgres is unreachable
```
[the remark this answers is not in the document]
nicolae536: Both halves fixed in f639b0d2.

`/ready`: the README now says 503 on Postgres being unreachable, and states explicitly that the custom-field schema is reported but never gates readiness — matching `_ready_response`, which keys `status_code` off `database_ok` alone.

Layout: `entity_types.py` is listed again.

services/backstop-mcp/src/backstop_mcp/server/tools/get_person.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+    if not isinstance(result, Resolved):
+        return tool_result(unresolved_party_response(result))
+
+    party = result.value
+    path = f"/people/{quote(party.id, safe='')}"
+    document = await client.get(
+        path,
+        params={"include": EntityRelationshipInclude.for_employment()},
+        schema=BackstopApiResourceDocument[PersonAttributes],
```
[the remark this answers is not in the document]
nicolae536: Fixed in d2e0770e — `get_person` now GETs `/{party.search_type}/{id}` so a contact/employee hit from the shared PERSON_* quick-search types uses the correct collection.

PR #756, opened 2026-08-07
fix(kb-mcp): persist OAuth state in Postgres and address PR #719 review feedback

proschu2:
## Summary

Targets `feat/kb-search` (PR #719) directly — this is the review follow-up for that PR, not a new feature.

**P0 fix:** `create_zitadel_oidc_proxy(...)` in `main.py` never passed `client_storage`, so FastMCP's `OAuthProxy` fell back to its own on-disk store under `settings.home / "oauth-proxy"`. That path doesn't exist in the deployed container — the Dockerfile creates the `app` user without `-m`, and the chart runs with `readOnlyRootFilesystem: true` (the base chart's own default; kb-mcp doesn't override `securityContext`) — so the unguarded `mkdir()` in that fallback raises before `mcp.run()`: CrashLoopBackOff. Even with a writable `HOME`, the same fallback is per-pod: the FastMCP access token is a reference token re-validated against this store on every request, so per-pod storage means every rollout 401s all live sessions with no recovery path, and >1 replica means a chunk of requests 401 depending on which pod served an earlier step.

Also addresses the ~22 review comments from `nicolae536` and `konsti` that motivated a broader cleanup alongside the fix (config consolidation had to land first so the storage fix's new env vars had one place to live).

## Changes (7 commits)

1. **refactor: consolidate configuration into one module** — replaces the scattered `ServerSettings`/`KbMcpServerSettings`/`_ContentTreeCacheSettings`/`_LOOKUP_CONCURRENCY` config surface (one of which read `ZITADEL_*` env vars from inside the `unique_mcp` package) with a single lazily-cached `kb_mcp.settings.Settings`. No env var names changed.
2. **fix: persist OAuth state in Postgres, fail closed** (the P0) — `kb_mcp/auth/storage.py` backs `client_storage` with an encrypted Postgres-backed key-value store (`py-key-value-aio[postgresql]` + Fernet). `Settings` refuses to boot on ephemeral storage unless `ALLOW_EPHEMERAL_OAUTH_STORAGE` is explicitly set for local dev.
3. **feat: chart support for Postgres and env-var mapping** — `postgresql.enabled: true` with a CloudNativePG connection example (same shape as `teams-mcp`/`outlook-semantic-mcp`); new `mcpConfig.{app.publicBaseUrl,app.frontendBaseUrl,zitadel.baseUrl}` values mapped to env vars via `templates/_ext.tpl`, replacing the hand-maintained 54-line env-var comment block; chart comments now describe how the Unique QA overlay actually delivers secrets (`extraEnvSecrets` + external-secrets).
4. **refactor: move modules next to their consumers** — each tool is now a folder (`tools/{search,content_tree,read_file}/{__init__,tool,config,...}.py`) instead of a flat module; verified `FileSystemProvider` still discovers all three tools from the nested layout.
5. **refactor: simplify auth and tool wiring** — `content_tree`/`read_file` tool descriptions moved into their docstrings (FastMCP infers them); deleted the CORS middleware (`/mcp` is server-side, never browser-called); `build_auth` now passes `required_scopes` and drops the separate `update_default_scopes` call — verified against fastmcp 3.4.5 that the `OIDCProxy` docstring's `invalid_token` warning is stale for `verify_id_token=True`.
6. **feat: add operational logging** — tool entry/exit/error logging with a truncated-hash correlation id (`kb_mcp/correlation.py`) instead of raw `user_id`/`company_id` (both are `SecretStr` in `unique_toolkit`'s own settings). Regression test per tool asserting no raw identifier appears in log output.
7. **test: assert behaviour, not internal calls** — rewrote the `content_tree` mode-dispatch tests and `search`'s service test to assert on the returned `CallToolResult` content instead of `mock.assert_called_once()`/`assert_called_once_with(...)`.

Commit 8 from the review plan (explicit `output_schema` for `search`) was dropped — optional, and the PR is already large enough per the review plan's own guidance.

## Deviations from the literal review discussion (with reasons)

- The new `Settings` module lives in `settings.py`, not `config.py` — `config.py` was already `SearchToolConfig` in this branch.
- `get_settings()` is a plain accessor, not a FastMCP `Depends()` — it's process-wide config with no per-request variance, so DI would be ceremony and would break direct-call unit tests.
- `search`'s tool description stays an explicit `description=` composed from `REFERENCE_FORMAT_INFORMATION` (also reused in `_META`) — a literal docstring can't reference a module constant, so moving it to the docstring would mean hand-duplicating that text with a real risk of drift.

## Known limitations

- No `helm` binary / OCI registry access in the environment this was built in (proxy-policy blocked) — the Helm chart changes were hand-validated against the merged `values.schema.json` via `jsonschema`, not `helm template`. `scripts/render-values-schema.sh --check` still needs a real run.
- No live Postgres/Zitadel available to smoke-test the P0 fix end-to-end (OAuth flow + client survives restart).
- 12 pre-existing `test_read_file_tool.py` failures are unrelated: tiktoken's BPE download is blocked by this sandbox's egress policy, verified identical on the unmodified base branch.

## Test plan

- [x] `uv run pytest` — 66 passed, 12 pre-existing/environmental failures (see above), 0 caused by this PR
- [x] `uv run ruff check .` / `uv run ruff format --check .`
- [x] `uv run basedpyright` — 0 errors
- [x] Confirmed `FileSystemProvider` discovers all three tools after the folder restructuring (ad-hoc script, not a committed test)
- [x] Hand-validated `values.yaml` + `ci-values.yaml` against merged `values.schema.json` via `jsonschema`
- [ ] `helm template` / `scripts/render-values-schema.sh --check` (needs `helm` + GHCR access)
- [ ] Manual smoke: OAuth + `/mcp` against a tenant, including a restart to confirm the client registration survives

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ho49gczJCq2ge76ypaV3Xz)_

---

Supersedes #753, which GitHub auto-closed when the branch was renamed from `claude/kb-mcp-p0-review-84us41`. Same commits, reauthored; the two resolved `cursor[bot]` threads from #753 are still readable there.


services/kb-mcp/src/kb_mcp/tools/search/tool.py
```diff
@@ -63,7 +59,12 @@
 
 @tool(
     name="search",
-    description=_TOOL_DESCRIPTION,
+    # Not the docstring: this reuses REFERENCE_FORMAT_INFORMATION (also in
+    # _META below), and a literal docstring can't reference a module constant.
+    description=(
+        "Search the knowledge base for the given query and return relevant "
+        "chunks. " + REFERENCE_FORMAT_INFORMATION
```
proschu2, line 66: here we need to use description since we cannot have variables in function docstrings

services/kb-mcp/src/kb_mcp/main.py
```diff
@@ -49,31 +32,18 @@ def main() -> None:
...
+    # No allowed_hosts: the DNS-rebinding guard is off by default in FastMCP
+    # 3.4.x, so passing it alone does nothing.
     mcp.run(
-        transport=server_settings.transport_scheme,
-        host=server_settings.local_base_url.host,
-        port=server_settings.local_base_url.port,
-        log_level="info",
+        transport=settings.transport_scheme,
+        host=settings.local_base_url.host,
+        port=settings.local_base_url.port,
```
nicolae536, line 46: shouldn't we pass log level from the config ?

services/kb-mcp/src/kb_mcp/settings.py
```diff
@@ -1,22 +1,132 @@
...
+
     def frontend_base_url_str(self) -> str | None:
         if self.frontend_base_url is None:
             return None
         return str(self.frontend_base_url).rstrip("/")
+
+    @model_validator(mode="after")
+    def _storage_must_be_durable(self) -> "Settings":
+        durable = bool(self.database_url and self.encryption_key)
+        half = bool(self.database_url) ^ bool(self.encryption_key)
```
nicolae536, line 98: funny I would have never used xor for this :D

nicolae536 (approved):

services/kb-mcp/src/kb_mcp/main.py
proschu2: fine with the env var, by the unique mcp code

PR #757, opened 2026-08-10
feat(teams-mcp): add independent CHAT_INTEGRATION toggle for composable capabilities

vabatta:
## What

Adds a new **`CHAT_INTEGRATION`** toggle (default `enabled`) that is **independent** of `UNIQUE_INTEGRATION`. The two "modes" are really two orthogonal capability axes, so instead of bolting a third value onto an enum we keep `UNIQUE_INTEGRATION` meaning exactly what it does today (ingestion/KB on/off) and add one new toggle for chat. This is fully backward compatible and lets an operator compose any combination.

This unblocks a new **least-privilege ingestion-only mode** (UN-23436): transcript/recording ingestion **without** any Teams chat/channel messaging permissions.

| Combination | `UNIQUE_INTEGRATION` | `CHAT_INTEGRATION` | ChatModule | KbIntegrationModule | Graph scopes |
|---|---|---|---|---|---|
| full | `enabled` | `enabled` | ✅ | ✅ | identity + messaging + KB |
| chat-only | `disabled` | `enabled` (default) | ✅ | ❌ | identity + messaging |
| **ingestion-only (new)** | `enabled` | `disabled` | ❌ | ✅ | identity + KB |
| both-off | `disabled` | `disabled` | ❌ | ❌ | → **fail fast at startup** |

Identity scopes (`openid, profile, email, offline_access, User.Read`) are always requested.

## Changes

**Core code**
- New `chat` config namespace (`CHAT_INTEGRATION`), `src/capabilities.ts` helpers, and a fail-fast startup assertion in `main.ts` when both capabilities are disabled.
- `microsoft.provider.ts`: split scopes into `IDENTITY_SCOPES` / `MESSAGING_SCOPES` / `KB_SCOPES`; `resolveMicrosoftScopes` and `createMicrosoftOAuthProvider` now take `{ chat, ingestion }`. `CHAT_SCOPES`/`SCOPES` kept as compositions for compat.
- `chat.module.ts`: `registerChatModule()` gating, mirroring `registerKbIntegrationModule()`.
- Both scope consumers (`app.module.ts`, `graph-client.factory.ts`) pass both toggles.

**Helm** — `chat.integration` value, unconditional `CHAT_INTEGRATION` env in `_helpers.tpl` (KB block untouched), both JSON schemas, README regenerated with helm-docs v1.14.2.

**Terraform module** — `chat_integration` variable (+validation) and `identity/messaging/kb` scope split via `setunion`.

**Docs & env** — permissions/configuration/tools/authentication/README/faq; `.env.example`, `.env.test`, `docker-compose.prod.yaml`.

**Tests** — provider spec (all 4 combinations), new `capabilities.spec.ts`, new `test/ingestion-only.e2e-spec.ts`; `get-token.ts` now reuses `resolveMicrosoftScopes`.

## Verification

- `tsc --noEmit` ✅ · `biome check` ✅ · unit tests **106/106** ✅
- Terraform `fmt`/`validate`/`console` confirm ingestion-only resolves to `User.Read` + 4 KB scopes (no messaging).

> **Note:** the e2e suite (incl. the new ingestion-only spec) needs live Postgres + RabbitMQ and was not run in the authoring environment; run it in CI/locally with infra up. Full `helm template`/`lint` needs the private `base` OCI chart.

## Backward compatibility

Existing deployments with `CHAT_INTEGRATION` unset behave exactly as before (chat enabled). The current infra module calls keep working unchanged.

## Follow-up

Infrastructure repo changes (Part B of the plan) are intentionally deferred — they must pin this PR's merge SHA, so they follow once this merges.

UN-23436


services/teams-mcp/src/chat/chat.module.ts
```diff
@@ -14,6 +15,15 @@ import {
   SendChatMessageTool,
 } from './tools';
 
+export function shouldRegisterChatModule(): boolean {
```
nicolae536, line 18: should we just have 1 function ?

nicolae536 (approved):

PR #773, opened 2026-08-12
feat(outlook-semantic-mcp,proxy,unique-api,utils,ci,deps): add forward proxy support

nicolae536:
## Summary
- Extract `@unique-ag/proxy` (`ProxyModule`, `ProxyService`, `PROXY_*` config schema) from the confluence-connector implementation, and route all nine outbound call sites in `outlook-semantic-mcp` through it — Graph SDK, token refresh, Unique API, health checks, and the attachment/blob upload paths.
- Proxy the `passport-microsoft` login flow, which the ticket missed: it runs on Node's core `https` agent and is unreachable by an undici dispatcher, so without this a proxied deployment could never complete a login. Handled via `oauth`'s `setAgent()` and `https-proxy-agent`, with no change to `@unique-ag/mcp-oauth`.
- Compose `interceptors.retry()` and `interceptors.redirect()` back onto the dispatcher passed to `UniqueApiModule`. `unique-api` only applies them when no dispatcher is supplied, so passing one would otherwise have silently dropped both — including when no proxy is configured.
- Add the Helm `proxyConfig` block, inject `PROXY_*` via `_config-proxy.tpl` / `_ext.tpl`, and chart tests, plus `.env.example` and operator documentation, matching the SharePoint connector's env var names and semantics.
- Delete the unused `HttpClientModule`, which composed retry/redirect on a non-proxied agent and would have become a silent proxy bypass.
- `sharepoint-connector` and `confluence-connector` are untouched; migrating them onto `@unique-ag/proxy` is tracked in [UN-24194](https://unique-ch.atlassian.net/browse/UN-24194).

## Ticket
UN-23554

## Test plan
- [ ] `pnpm --filter @unique-ag/proxy test`
- [ ] Outlook unit tests for Graph factory, token provider, Unique API compose, microsoft provider, health indicator
- [ ] Helm chart tests for `proxyConfig` (disabled / none / no_auth / username_password / ssl_tls)
- [ ] Smoke with `PROXY_AUTH_MODE=none`: behaviour unchanged
- [ ] Smoke with `no_auth` or `username_password` proxy: Graph login, token refresh, and Unique API succeed through the proxy when `UNIQUE_SERVICE_AUTH_MODE=external`

[UN-24194]: https://unique-ch.atlassian.net/browse/UN-24194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

https://github.com/user-attachments/assets/36c91d7e-d44b-4453-9ea2-290ee5b7e92d


packages/proxy/src/proxy.config.ts
```diff
@@ -0,0 +1,72 @@
...
+const noAuthProxyConfigSchema = z.object({
+  authMode: z.literal('no_auth').describe('Proxy enabled without authentication'),
+  ...baseProxyFields,
+});
+
+const basicProxyConfigSchema = z.object({
+  authMode: z.literal('username_password').describe('Basic authentication'),
+  ...baseProxyFields,
+  username: requiredStringSchema.describe('Proxy username'),
+  password: redacted(z.string().nonempty()).describe('Proxy password'),
```
Ostrzy: Don't we want to use the same approach that is in confluence connector where this can be defined as `os.environ/ENV_VAR` to point to an arbitrary env var? 🙂

packages/proxy/src/proxy.config.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const noAuthProxyConfigSchema = z.object({
+  authMode: z.literal('no_auth').describe('Proxy enabled without authentication'),
+  ...baseProxyFields,
+});
+
+const basicProxyConfigSchema = z.object({
+  authMode: z.literal('username_password').describe('Basic authentication'),
+  ...baseProxyFields,
+  username: requiredStringSchema.describe('Proxy username'),
```
Ostrzy: For one of the client, they didn't like username being plaintext in the helm configs or anything like that, so it should probably not go in logs as well :) The same as done for COnfluence COnnector: https://github.com/Unique-AG/connectors/pull/774/changes

Ostrzy (approved):

packages/proxy/src/proxy.config.ts
nicolae536: Done

packages/proxy/src/proxy.config.ts
nicolae536: Done

services/outlook-semantic-mcp/package.json
```diff
@@ -11,7 +11,7 @@
   ],
   "scripts": {
     "build": "nest build",
-    "dev": "nest start --watch",
+    "dev": "NODE_EXTRA_CA_CERTS=/Users/nicolaebacila/.mitmproxy/mitmproxy-ca-cert.pem nest start",
```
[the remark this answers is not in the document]
nicolae536: Fixed

packages/proxy/src/proxy.config.ts
```diff
@@ -1,10 +1,25 @@
...
+  if (!val.startsWith(ENV_REF_PREFIX)) {
+    return val;
+  }
+  const varName = val.slice(ENV_REF_PREFIX.length);
+  return process.env[varName] ?? '';
+});
+
+const envRequiredSecretSchema = envResolvableStringSchema
+  .pipe(z.string().trim().nonempty())
+  .transform((val) => new Redacted(val));
```
Ostrzy: On a tiny note, `envRequiredSecretSchema` and `envResolvableStringSchema` could go to zod utils in the utils lib :slightly_smiling_face:

Ostrzy (approved):

packages/proxy/src/proxy.config.ts
nicolae536: Fixed

PR #774, opened 2026-08-13
fix(confluence-connector): resolve basic auth username from env references

lorand93:
Basic auth `username` only accepted a plain string, so `os.environ/VAR` was passed through literally and authentication failed. It now uses `envRequiredSecretSchema`, the same env-resolvable, log-redacted schema as the other credentials.

Also trims surrounding whitespace from credentials, since secrets injected from Kubernetes secrets or `.env` files routinely carry a trailing newline that Confluence rejects with a bare 401.

### DIFF
--- a/services/confluence-connector/src/config/confluence.schema.ts
+++ b/services/confluence-connector/src/config/confluence.schema.ts
@@ -25,7 +25,7 @@ const patAuth = z.object({
 
 const basicAuth = z.object({
   mode: z.literal(AuthMode.Basic).describe('HTTP Basic username/password (Data Center only)'),
-  username: requiredStringSchema,
+  username: envRequiredSecretSchema,
   password: envRequiredSecretSchema,
 });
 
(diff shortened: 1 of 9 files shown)


services/confluence-connector/src/config/confluence.schema.ts
```diff
@@ -25,7 +26,7 @@ const patAuth = z.object({
 
 const basicAuth = z.object({
   mode: z.literal(AuthMode.Basic).describe('HTTP Basic username/password (Data Center only)'),
-  username: requiredStringSchema,
+  username: envRequiredPlainSchema,
```
lorand93: where: 
```
export const envRequiredPlainSchema = envResolvableStringSchema.pipe(z.string().trim().nonempty());

const envResolvableStringSchema = z.string().transform((val) => {
  if (!val.startsWith(ENV_REF_PREFIX)) {
    return val;
  }
  const varName = val.slice(ENV_REF_PREFIX.length);
  return process.env[varName] ?? '';
});

```

Ostrzy (approved):

lorand93:
bugbot review

PR #778, opened 2026-08-13
feat(office-mcp): call Microsoft Graph as the signed-in user

vabatta:
## Summary

Adds `graph_client/`, the Microsoft Graph transport that tools call through. The package wraps the official `msgraph-sdk` and handles token binding, error classification, and paging.

The transport is one shared `httpx.AsyncClient` (connection pool plus SDK middleware). Each Graph call gets its own token via `graph_client_for()`. FastMCP's On-Behalf-Of exchange hands the caller's delegated token as a string; the transport attaches it as a bearer header to graph.microsoft.com only.

Failures are classified into four categories with remedies: throttled (429, retriable), forbidden (401/403), not-found (404), unavailable (5xx). The SDK's retry middleware handles `Retry-After` waits; this package returns typed errors when retries are exhausted.

Paging follows `@odata.nextLink` via the SDK's `PageIterator` with two caps: item limit and scan limit. A `truncated` flag says whether the answer may be incomplete. Search tools do not use this module — they use stateless offsets instead.

Timeouts are sized for interactive MCP calls (30 s request, 10 s connect) instead of the SDK's batch defaults. `GraphSettings` is a frozen dataclass; `create_app` translates app config into it, and nothing imports from `office_mcp.config` here.

Tests cover token binding, throttling and retries, error classification from status codes, and pagination with caps. Layering rules are enforced: `graph_client/` imports neither `features/` nor `config`.

No MCP tools are registered yet. The service still boots and answers health and ready probes with no tools.

## Role in the stack

This PR provides the Microsoft Graph transport — the HTTP pipeline and error handling that every later Teams feature tool calls through. It builds on the Entra auth PR, which established how the service acquires and caches the caller's delegated tokens. No tool PR can merge until this transport exists. The next PR adds the first tool, `get_me`, atop this transport.


vabatta:

PR #779, opened 2026-08-13
feat(office-mcp): expose get_me over Microsoft Graph

vabatta:
Add the first MCP tool, `get_me`, and the seam every later tool will inherit: a per-tool token dependency, error mapping that names missing permissions, and a shared read-only annotation. It also introduces the knob an operator uses to decide which tools a deployment runs at all — see **Tool surface** below.

`get_me` uses `GET /me` under the `User.Read` permission and returns the signed-in user's profile: `user_id`, `display_name`, `email`, `user_principal_name`, and `job_title`. Field names are snake_case everywhere on this connector, and two of them depart from Graph's own spelling on purpose — `user_id` (Graph `id`) and `email` (Graph `mail`) are the two identifiers the tool description warns are not interchangeable, so they get names that say what they are rather than echoing Graph. The Graph call itself lives in `shared/identity.py`, separate from the tool file, because "who is the signed-in user" is a fact later tools will also need (to compare against a meeting organizer, say) and a second `GET /me` under a different projection would be a second answer to one question.

`shared/seam.py` is new and is the one file under `shared/` allowed to import FastMCP. `GraphToken` wraps FastMCP's `EntraOBOToken` dependency so that an On-Behalf-Of exchange Entra refuses becomes a `ToolError` naming the missing permission, instead of FastMCP's own dependency-resolution error, which names neither. `graph_tool_errors` maps a Graph failure to advice from the same shape every tool uses: 401 asks the user to sign in again, 403 names the permission an administrator must grant, 429 carries Graph's `Retry-After`, and diagnostics (status, Graph error code, request id) are appended once, only where Graph sent them. `READ_ONLY` is the MCP annotation every read-only tool on this connector declares.

`tools/__init__.py` is the registry. A tool module publishes `TOOL_NAME`, `GRAPH_PERMISSIONS` and `register`, and the registry turns an operator's selection into both halves of one answer: the modules to register, and the union of their permissions as the scope list `create_app` hands the auth provider in `additional_authorize_scopes`. Both are derived from the registered modules via `dict.fromkeys`, so the scope list is never hand-maintained and never drifts from what a tool actually declares. `create_app` resolves once and hands the same `Selection` to `build_auth` and to `register_tools`, so the two cannot disagree about a tool.

## Tool surface

Exactly one of `TOOLS_PRESET` and `TOOLS_ENABLED` says which tools a deployment runs, and the permissions sign-in asks every user to consent to are exactly the union of what those tools need. Without it, every deployment asked every tenant for every tool's permissions: an operator who wanted one part of this connector still needed an administrator to consent to the rest, and if the administrator refused, Entra rejected the authorize request and **every sign-in failed for every user**, for a tool nobody called.

There is deliberately **no default**. `TOOLS_PRESET=teams` is "every tool there is" — a one-word but *chosen* value — because a default would make the widest consent screen what an operator gets by not choosing. Neither variable set is a startup error; both set is a startup error naming which to remove.

- **The names of the presets live in `config.py`**, beside `AppEnv` and `LogLevel`, so a misspelling is a startup error listing the values that would have worked, and the Helm schema can carry the same `enum`. **What each name expands to stays in the registry**, the only module that knows which tools exist — `config.py` importing tool names would invert the dependency and be a second place the tool list lives. A test asserts the two sides agree in both directions.
- **The filter happens in the registry, before the union is taken.** FastMCP's `enable`/`disable` visibility transforms would leave a module registered and its scopes already computed, so they shorten `tools/list` and change nothing whatever about what the tenant is asked to grant.
- **Registry order, never the operator's.** `TOOLS_ENABLED=a,b` and `b,a` must produce the same scope list, because the consent screen and every cached On-Behalf-Of token key are keyed by it as a string.
- **`get_me` is an always-on floor.** It is how this server resolves "me", and `User.Read` is the least-privileged delegated permission Microsoft publishes and needs no administrator. So `TOOLS_ENABLED` lists only the rest, presets need not name it, naming it anyway is accepted rather than an error, and no deployment asks for zero permissions.
- **Every wrong way of writing a selection aborts startup naming the remedy**: an unknown tool name, an unknown preset, an empty list, both variables, neither. A typo must never quietly cost a tool — it would register one tool fewer and ask for one permission fewer than its operator believes, and the first sign of it would be a model that cannot find a tool the deployment was supposed to expose.

**No tool file changes.** `TOOL_NAME` and `GRAPH_PERMISSIONS` were already on every tool file; the `ToolModule` protocol simply reads the first as well as the second.

### The manifest

A startup log line and `GET /manifest` print what was resolved: the tools, the exact delegated permissions in Entra's spelling, and which of them need admin consent. That is the only place the exact ask is written down. Azure omits Graph scopes from the session token's `scp`, so this server cannot check its own ask against the app registration — and a scope the registration does not carry fails at the **authorize** hop, for every user, with nothing in these logs. Those lines are what an operator hands their Entra administrator.

It prints **no consent URL**: provisioning the registration is out of scope, `/.default` would consent to whatever the registration happens to carry rather than to what this deployment asks for, and a scope-matched admin-consent URL would need a second registered redirect URI to land on.

Riding along is a description scan reporting prose that points a model at a tool this deployment does not expose. It only **warns**. Tool prose names its siblings densely and mutually, so requiring every mention would drag `ChannelMessage.Read.All` — an administrator's signature — into a deployment that wanted nothing but a chat list.

### Deployment

`.env.example` carries a selection so `uv run office-mcp` still works out of the box. The value rides the chart's existing `env:` map; `values.yaml` deliberately defaults neither variable, because a chart-level default would restore the maximal ask through the back door, and `values.schema.json` requires exactly one of the two and carries the preset names as an `enum` — so a missing or misspelled selection fails `helm install` at validation instead of crash-looping a pod. Widening a live selection is recorded beside the client-secret rotation note, because it costs the same thing: every signed-in user signs in again.

## Layering

The `features/` package from the scaffolding PR is deleted; the layout is now `shared/` (vocabulary two tools must not disagree about), `tools/` (one file per tool, plus the registry), and `server/` (the `/ready` probe and the manifest — exposure that is not a tool). `test_layering.py` is rewritten around this layout: `shared/` may import no tool module and only `seam.py` may import FastMCP; `graph_client/` may import nothing of the application; `tools/` may import only `shared/`, `graph_client/`, and FastMCP; a package is entered through its `__init__`; and only `create_app` may construct a config class — which `SurfaceConfig` now joins. Three further rules are named but not yet assertable with one tool registered — no tool module may import another, one speller per handle family, and no module may address a single meeting recording — each arrives with the PR that makes it real.

## Role in the stack

This PR adds the first tool, the seam every later tool inherits, and the operator-facing tool surface. It builds on `office-mcp/feat/graph-client` (#778), which provided the transport. `list_chats` (#797) is the next tool stacked on this one, and reuses the seam rather than rebuilding it. The curated use-case presets land at the top of the stack, in `office-mcp/feat/recordings` (#786), which is the first point at which every use case they name exists. This PR is independently reviewable and mergeable.


vabatta:

PR #780, opened 2026-08-13
feat(office-mcp): search Teams messages over the Graph Search API

vabatta:
This PR adds the `search_messages` tool, the flagship read tool for the office-mcp service. The tool searches Teams messages using `POST /search/query` with `entityTypes: ["chatMessage"]`, the only full-text path Microsoft Graph offers over Teams messages.

The tool requires two Graph permissions: `Chat.Read` and `ChannelMessage.Read.All`. This is the first tool needing two permissions. When Graph or Entra refuses the request, the error message names both permissions, since neither Graph's 403 nor Entra's token-exchange failure says which permission was missing.

Search results are projections, not full messages. Each result carries metadata (sender, timestamp, ids), a `summary` snippet from Microsoft's index, and a `uri` handle. A caller reads the full message body by passing the uri to a message-reader tool in a later PR.

Key design constraints:
- At least one search criterion is required. The tool enforces this at the tool boundary with JSON Schema `anyOf` plus a runtime check, since FastMCP validates against the function signature rather than the advertised schema.
- The tool uses stateless paging with `from` and `size` integers, not an opaque cursor. A search resumes across separate calls.
- No total match count is reported. Graph returns only the count on the current page. The tool reports `more_results_available` and `next_offset` only.
- No per-chat scan fallback. The prior connector fell back to scanning up to 50 chats per request when permissions were missing. Graph's read budget is one request per second per app per tenant on a given channel or chat. The *per app* part means one user's scan degrades every other user in the tenant. Date bounds go into the query string instead, where Microsoft's index applies them.

The free-text query parameter receives a word-by-word KQL injection guard. A multi-word query is passed as separate terms, which KQL ANDs together. A caller who wants phrase matching quotes the words themselves. This differs from the prior connector, which quoted the entire query into a phrase, silently dropping all results whose words were not adjacent.

Query text is never logged or stored in a span attribute, since what someone searched their messages for names people and deals. A test asserts this.

## Role in the stack

This PR adds the message search capability, using `POST /search/query` and minting `teams:///...` message handles. It is the first tool needing two Graph permissions.

The PR builds on #797 (list_chats), which established `shared/handles.py` as the one place that spells or parses a `teams:///` URI — this tool adds the message handle family to it rather than minting its own.

The next tool, a message reader, will resolve these handles into full message bodies. That tool is stacked on this one and requires the same `ChannelMessage.Read.All` permission.


vabatta:

PR #782, opened 2026-08-13
feat(office-mcp): read Teams messages from search handles

vabatta:
Search results carry metadata and a `teams:///` handle but no body — Graph's search index returns a reduced chatMessage projection. This PR adds `read_message` to resolve handles and return the full message.

The tool accepts the two handle shapes Graph uses: `teams:///chats/{chatId}/messages/{messageId}` for chats and `teams:///teams/{teamId}/channels/{channelId}/messages/{messageId}` for channels.

The message text is normalized from Teams HTML: mentions become `@Name`, list items become `- `, emoji are kept as themselves, inline images become `[image]`, attachments become `[attachment: name]`, and cards become `[card]`.

System event messages report the event type from `eventDetail.@odata.type` rather than the literal `<systemEventMessage/>` tag. Deleted messages report `deleted_at` and no text.

The sender uses the same normalization as search, so a mailbox-shaped emailAddress and a teamworkUserIdentity report one sender.

Cards are detected from attachment `contentType`, not from the shape of body text. This fixes a bug where the prior heuristic would discard any message somebody pasted JSON into — a config fragment, an API response, a log output — and report it as a card with its content thrown away.

## Role in the stack

`read_message` gives the full message body that search results cannot provide. It follows the PR that adds `search_messages`, which returns metadata plus a handle. Together they let a model answer "what did they actually say" from a search result.

Messages can now be read in full. Transcripts and other M365 resources get their own handle-taking readers in later PRs.


vabatta:

PR #783, opened 2026-08-13
refactor(office-mcp): make the four tools read as one server

vabatta:
Wave 1 integration. This pull request makes the four tools `get_me`, `list_chats`, `search_messages`
and `read_message` read as one server. **It adds no capability.** It deletes, renames, unifies and
rewords.

One commit, 12 files, +420/-287. The last section lists the gaps that I found and did not build.

## Where this branch sits in the stack

The base branch is `office-mcp/feat/read-resource`, which is pull request #782. The stack starts at
pull request #775, and #775 is the only one of them that targets `main`. `main` holds no
`services/office-mcp` directory today. This diff assumes that #782 merges first. A change under #782
needs a rebase here.

## The renames, and what they break

| Old | New | What it names |
| --- | --- | --- |
| `whoami` | `get_me` | the tool name |
| `Who Am I` | `Get My Profile` | the title of that tool |
| `List My Chats` | `List My Teams Chats` | the title of `list_chats` |
| `id` | `user_id` | a field of `SignedInUser`, the output of `get_me` |
| `mail` | `email` | a field of `SignedInUser`, the output of `get_me` |
| `more_results_available` | `truncated` | a field of `MessageSearchResults`, the output of `search_messages` |

The declared output schema of `get_me` and the declared output schema of `search_messages` both
change, and `tests/test_mcp_tools.py` asserts the new ones. A client that reads `whoami`, `id`,
`mail` or `more_results_available` breaks. `git grep` finds no reference to `office_mcp` or to
`more_results_available` outside `services/office-mcp`. It finds one reference to `whoami` outside
it: `op whoami` in a deploy script of temenos-mcp, which is the 1Password CLI.

## What the committed tests cover

`tests/test_mcp_tools.py` drives the server in process, over the MCP protocol. It uses
`fastmcp.Client` against the application that `create_app` builds. respx stubs Graph, and
`_StubOboCredential` takes the place of the Entra On-Behalf-Of credential and records the scopes
each call asks for. The list below names the tests that I read, with the file of each one. No test in
`tests/test_mcp_tools.py` calls more than two of the four tools in one sequence.

- `test_a_handle_from_a_search_result_reads_the_message_behind_it`, in `tests/test_mcp_tools.py`,
  calls `search_messages`. It then passes the `uri` of the first hit to `read_message` unchanged.
  That handle is `teams:///chats/19%3Arelease%40thread.v2/messages/1770000000000`. The read answers
  with the text, one mention and the display name of the sender.
- One model, `MessageSender`, serves both message tools. Its three fields are `display_name`,
  `email` and `user_id`. One function, `sender_of`, builds it from both identity shapes of Graph.
- The chat id `19:release@thread.v2` is asserted from two sides. `tests/test_mcp_tools.py` asserts it
  on a `list_chats` result, and `tests/features/test_message_search.py` asserts it on a search hit.
  No test compares the two values inside one sequence of calls. `chat_id` is an argument to no tool,
  and the descriptions now say so.
- The OBO exchange asks for the scopes of the called tool. `get_me` asks for
  `https://graph.microsoft.com/User.Read`. `list_chats` asks for
  `https://graph.microsoft.com/Chat.Read`. `search_messages` and `read_message` both ask for
  `Chat.Read` and `ChannelMessage.Read.All`. `read_message` asks for both because the exchange runs
  before the tool parses the handle.
- `TestWhatAModelIsToldWhenGraphRefuses`, in `tests/test_mcp_tools.py`, covers seven failures at the
  MCP boundary:
  - 403 on `GET /me/chats`, which names `Chat.Read` and an administrator.
  - An AADSTS65001 refusal before Graph on `list_chats`.
  - The same refusal on `search_messages`, which names both permissions.
  - Three arguments that are not handles, in one parametrized test.
  - 404 on a handle with a correct shape.
  - 403 on a chat read, which names `Chat.Read` alone.
  - 403 on a channel read, which names `ChannelMessage.Read.All` alone.

  Each of those tests asserts the remedy in the message, and not only the failure. The 403 on
  `GET /me/chats` and the 404 both assert the Graph request id of the stub.
- `tests/server/test_errors.py` holds unit tests over `server/errors.py`. Those cover 429 with
  `Retry-After`, 429 without it, 500, 503 and 400. No test drives those statuses through the MCP
  boundary.

## The process, run by hand

I ran the process once by hand, and this run is no part of the gates. A reviewer has to repeat it.
The command is `uv run office-mcp` in the nix dev shell. The settings are `APP_ENV=development`,
`PORT=9599` (the committed default is 9544), `PUBLIC_BASE_URL=http://localhost:9599` and the three
`ENTRA_*` values. The database is a Postgres on `localhost:9001` that holds `office_mcp`.

| Route | Result |
| --- | --- |
| `GET /health` | 200 `{"status":"healthy"}` |
| `GET /probe` | 200 `{"status":"ok"}` |
| `GET /ready` | 200 `{"status":"healthy","checks":{"database":true}}` |
| `GET /metrics` | 200, Prometheus text (`mcp_calls_total`, `python_http_requests_total`) |
| `GET /.well-known/oauth-authorization-server` | 200, `issuer`, `authorization_endpoint`, `token_endpoint`, `registration_endpoint`, `scopes_supported: ["access_as_user"]`, `code_challenge_methods_supported: ["S256"]` |
| `GET /.well-known/oauth-protected-resource/mcp` | 200, `resource` and `authorization_servers` |
| `POST /mcp` with no token | **401** with `www-authenticate: Bearer resource_metadata="http://localhost:9599/.well-known/oauth-protected-resource/mcp"` |

## The differences I found, and how this pull request resolves them

1. **Tool names.** Three tools used `verb_noun`, and `whoami` is a shell idiom on the first tool a
   model calls. The new name is **`get_me`**. `server/tools.py` now records the convention, and
   records that the M365 connector of Microsoft arrived at the same name. The new test
   `test_the_whole_surface_speaks_one_language` asserts three things against the live schemas: every
   tool name matches `[a-z]+(_[a-z]+)+`, every output field is snake_case, and `truncated` is an
   output field of `list_chats` and of `search_messages`. A rename breaks the expectations of a
   model, so the rename happens before more tools arrive.

2. **The words for pagination.** `list_chats` said `truncated` and `search_messages` said
   `more_results_available`. Both now say **`truncated`**. Each description states how to get the
   rest: a wider `limit` where there is no cursor, `next_offset` where there is one. `next_offset`
   stays, because it counts the hits of Graph itself. Those hits include the system messages that
   the tool drops, so a model cannot calculate the value.

3. **Field names.** The identity tool returned `id` and `mail`, and every other payload calls these
   things `user_id` and `email`. The new names are **`user_id`** and **`email`**, and the descriptions
   still name `id` and `mail` of Graph. The docstring of `SignedInUser` already claimed this rule,
   and the rule now holds.

   Timestamps were already uniform: every timestamp field ends with `_at`. `last_modified_at` on a
   search hit and `last_edited_at` on a read are two separate fields, and the description of
   `last_edited_at` names the property `lastEditedDateTime` of Microsoft. Both fields stay, and the
   description of `last_modified_at` now points at `last_edited_at`.

4. **The handle had two implementations.** `message_search` built `teams:///…` with its own
   `_segment`. `message_read` built the same two shapes a second time in `MessageHandle.uri`. The two
   permission constants, `MessageHandle`, `message_handle` and `_handle` moved into
   **`message_search`**. That module already had a `_segment` of its own, and it is the one module
   that mints a handle. The docstring of `message_read` already named that module as the owner of the
   contract.

   `message_read` imports them, as it already imported `sender_of`. Under `src/`, the literal
   `"Chat.Read"` drops from three places to two, and `"ChannelMessage.Read.All"` from two to one.
   The search keeps one small helper, renamed from `_uri` to `_hit_uri`, which builds a
   `MessageHandle` and returns its `uri`. The test class for the handle moved from
   `test_message_read.py` to `test_message_search.py`, and its name changed from
   `TestTheHandlesItAccepts` to `TestTheHandleItMints`.

5. **The words in an error message.** `_remedy` in `server/errors.py` writes seven messages. Six of
   them already opened with "Microsoft 365". The seventh is the fallback: it opened with "Microsoft
   Graph" and now opens with "Microsoft 365". The top of the module now states the shape every
   message follows. The actor comes first, then the remedy and whether a retry can help, then the
   diagnostics in brackets at the end.

   `_diagnostics` appends `HTTP`, `Graph error code` and `Graph request id` for the values Graph
   sent, on every message built from a Graph failure. The refusal from Entra is no Graph failure:
   `_token_advice` writes it, and it ends `(Entra AADSTS65001)` with no Graph request id. The diff
   changes no other message in `server/errors.py`.

6. **A description promised a tool that does not exist.** `list_chats` said "use it to find the
   `chat_id` that a chat-scoped tool needs". No tool here takes a chat id. A model that follows that
   sentence looks for a missing tool, or it assembles a handle, which `read_message` refuses.

   The description now states what `chat_id` is for: it names the chat that a search hit came from.
   It also states that no tool takes a chat id, and that a search cannot narrow to one chat.
   `search_messages` now states the same from its own side. `list_chats` also states that it returns
   no message text, and that `search_messages` is the only route to message text.

7. **Layering rules.** All eight rules ran before this pull request, and all eight run now.
   `tests/test_layering.py` carried three `skipif` markers, and this pull request removes all three.
   Two of them sat on the `graph_client` classes and were already inert, because `graph_client/`
   holds `__init__.py`, `client.py`, `errors.py`, `pagination.py` and `settings.py`. The third
   skipped `test_the_feature_tree_is_actually_there`, the guard over rule 1, and that was the one
   skipped test on the base branch.

   The rewritten guard asserts that `features/` holds one module or more that is not `__init__.py`.
   The dead `_feature_packages()` helper is gone.

8. **The README.** The renames reach the intro, the permission table and the tool table. It gains a
   paragraph on the ownership of the handle and of the sender shape, and a bullet on the conventions
   of the four tools. The bullet on `list_chats` now states what a `chat_id` is for. The bullet on a
   refusal now states the one error voice. The paragraph on the layering rules now states that each
   rule has a guard. It also states that no guard waits on a package that has yet to land.

## Deployability

The diff touches no file under `services/office-mcp/deploy`, so this round changed no configuration
surface. Nothing in the environment configures `GraphSettings`: it is a frozen dataclass of default
values, and `create_app` calls `GraphSettings()` and injects the result. I verified two points rather
than an assumption:

- `helm dependency update && helm template . -f ci-values.yaml` renders with exit status 0. The
  `validate-helm` job of `.github/workflows/python-ci.yaml` runs those two commands. The rendered
  manifests hold `APP_ENV`, `LOG_LEVEL`, `PORT` and `PUBLIC_BASE_URL`, and the `/health` and `/ready`
  probes.
- `scripts/render-values-schema.sh --check` reports OK for office-mcp and for six other charts.

## What this pull request does not change

- **`members_may_be_incomplete`.** I did not rename it to `truncated`. It makes a different claim
  about a different thing: a cap inside one chat cut the member list short. A model that must answer
  "who is in this chat" reads the explicit name. It does not trust the list.
- **A `GRAPH_PERMISSION` constant in each module.** `chats.py` still spells `Chat.Read` itself. A
  stated rule gives each feature the permission that its own request needs, next to the call that
  requires it. The messages pair collapsed only because the *handle* forced it.
- **`matches` and `max_scanned` on `collect_pages`.** The one caller under `src/` passes neither, and
  only `tests/graph_client/test_pagination.py` exercises them. They are the lesson that the
  pagination part imported from teams-mcp on purpose: a filtered collection can read many pages and
  keep few items. A deletion drops that lesson before the first filtered collection arrives.
- **The Helm chart validates `PUBLIC_BASE_URL` and does not validate `ENTRA_*`.** A new check is new
  chart behaviour. The localhost issuer is a different case: the container restarts in a loop and
  nothing states the cause. `.env.example` records that this service does not start without all
  three `ENTRA_*` values.
- **The privacy rule.** This pull request does not touch it, and two tests still assert it. No
  message text and no query term reaches a log record or a span attribute, over the whole call. The
  test for a read is `test_the_message_text_reaches_the_caller_and_no_log_or_span`. The test for a
  search is `test_the_search_text_reaches_graph_and_nothing_else`.

## Gaps for the owner, found and not built

- **A model cannot scope `search_messages` to a chat, a channel or a team.** The tool has no such
  parameter. The descriptions now state that, in place of an implication of the opposite. "What did
  we discuss in *this* chat" is the obvious question that a model asks. The owner must make a
  decision.
- **`ChatMember` carries `display_name` and `email` only.** Without an id, a model cannot answer
  "which participant is the signed-in user" from an id. It can match on the display name, or on the
  email address when the caller sets `include_member_emails`. The descriptions of `get_me` and
  `list_chats` now admit this.
- **No tool reads the messages of a chat by id.** `list_chats` gives a `chat_id`, and a model can
  correlate it against the output of search only. A tool of the shape of `list_messages`, or a chat
  scope on search, is what this pair still needs.
- **`read_message` cannot address a reply inside a channel thread.** The advice on a 404 states that
  such a reply is addressed under its parent post, and that this handle shape cannot express it.

## Gates

All four gates pass on this branch.

- `uv run pytest`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run basedpyright .`

### DIFF
--- a/services/office-mcp/README.md
+++ b/services/office-mcp/README.md
@@ -2,9 +2,9 @@
 
 An MCP server over Microsoft 365, via the Microsoft Graph API.
 
-Users sign in with their own Microsoft account and the server acts as them. It exposes three tools
-so far — `whoami`, `list_chats` and `search_messages` — and more land in later PRs, stacked on top
-of this one.
+Users sign in with their own Microsoft account and the server acts as them. It exposes four tools
+so far — `whoami`, `list_chats`, `search_messages` and `read_message` — and more land in later PRs,
+stacked on top of this one.
 
 ## Layout
 
@@ -15,7 +15,7 @@ src/office_mcp/
   auth.py                Entra auth: which app registration, and where its state lives
   logging.py metrics.py  cross-cutting, used by both sides below
   graph_client/          the Microsoft Graph transport — the official SDK, one caller's token
-  features/              what the connector does — one module per slice (identity, chats, search)
+  features/              what the connector does — one module per slice (identity, chats, messages)
   server/                how it's exposed over MCP — the tools, the errors they report, /ready
 ```
 
@@ -30,8 +30,9 @@ The layering rules are that **nothing under `features/` may import from `server/
 wires features together, never the reverse — nor **from FastMCP**, since deciding what an MCP client
 is told (a `ToolError`, a schema, an annotation) is the tool layer's job; that **`graph_client/`
 imports neither `features/` nor `config`**, taking its own frozen `GraphSettings` instead, that
-**`server/` does not import the Graph SDK**, since a tool declares and exposes while the request
-belongs to a feature, and that
+**`server/` does not import the Graph SDK** — nor the `kiota_*`/`msgraph_core` request layer
+underneath it, which is how a Graph call gets shaped without ever spelling `msgraph` — since a tool
+declares and exposes while the request belongs to a feature, and that
 **only `create_app` constructs a config**, so nothing downstream can quietly re-read the
 environment and disagree with the app it runs in. `tests/test_layering.py` enforces all of them,
 and grows as each package arrives.
@@ -75,8 +76,8 @@ consented to, which is why sign-in has to ask for it:
 | Permission | Type | Admin consent | Used by |
 | --- | --- | --- | --- |
 | `User.Read` | Delegated | No | `whoami` |
-| `Chat.Read` | Delegated | No | `list_chats`, `search_messages` |
-| `ChannelMessage.Read.All` | Delegated | Yes, in most tenants | `search_messages` |
+| `Chat.Read` | Delegated | No | `list_chats`, `search_messages`, `read_message` (chats) |
+| `ChannelMessage.Read.All` | Delegated | Yes, in most tenants | `search_messages`, `read_message` (channels) |
 
 `Chat.Read` rather than the least-privileged `Chat.ReadBasic` because listing chats by recency needs
 `$expand=lastMessagePreview`, and a message preview is a message.
@@ -86,7 +87,9 @@ enough for Graph to *accept* a `chatMessage` search, but Microsoft documents tha
 returns more than the equivalent GET would, and every channel-message GET in v1.0 requires
 `ChannelMessage.Read.All` — so without it a search silently covers chats only and reports nothing
 missing. Asking for it at sign-in makes a tenant that withholds it fail visibly at consent rather
-than serve half an answer per query. It is also what the message reader in the next PR needs.
+than serve half an answer per query. It is also what `read_message` needs for a channel message —
+Graph's permissions for a message read are per surface, so that tool's own 403 names whichever one
+the handle it was given required.
 
 **State.** Every token the server issues is a reference token re-validated on each request, so
 where that state lives decides whether the deployment survives a restart or a second replica.
@@ -132,6 +135,7 @@ gets that string onto the wire.
 | `whoami` | Who the signed-in user is: `id`, `display_name`, `mail`, `user_principal_name`, `job_title` | `GET /me` |
 | `list_chats` | The user's Teams chats, most recently active first | `GET /me/chats` |
 | `search_messages` | Teams messages matching keywords, a sender, mentions, a date range, attachment or read state | `POST /search/query` |
+| `read_message` | One Teams message in full — text, sender, mentions, attachments, edits, deletion | `GET /chats/{id}/messages/{id}`, `GET /teams/{id}/channels/{id}/messages/{id}` |
 
 All are read-only, all take their caller's identity from the token rather than from a parameter, and
 all are described to the model in prose that names the traps rather than only the fields — `mail` is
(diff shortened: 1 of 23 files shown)


vabatta:

PR #784, opened 2026-08-13
feat(office-mcp): list the teams the signed-in user is a member of

vabatta:
Add the `list_teams` tool to office-mcp. It lists the Microsoft Teams teams the user is a member of.

This is the first step onto the channel side of Teams. A chat is a conversation the user participates in. A channel is one the user can browse. Nothing reaches a channel without first naming the team it lives in.

The tool uses the `Team.ReadBasic.All` permission, which is the least-privileged delegated scope Microsoft documents for this endpoint. A tenant that withholds the broad channel message permission can still list its teams.

The tool sends no OData query parameters. Microsoft Graph does not support any OData parameters on the `/me/joinedTeams` endpoint, so the tool sends only the path. The tool applies `limit` as a window over the pages Graph returned. As many teams as `limit` means the user may have more teams. Fewer than `limit` is all of them.

Graph returns only these properties on this endpoint: id, displayName, description, isArchived, and tenantId. The tool reports all of them. No member count, no channel list, and no activity date is claimed.

Update the permission table in README.md to document the new scope.

Add tests for the tool and for the permission in sign-in.

## Role in the stack

This PR adds the `list_teams` tool to the office-mcp server. It is the only tool that reaches the team side of Teams — all other tools work on chats or messages within channels. The tool is not the channel list itself. The tool gives a model the team id needed to name where a channel hit lives. The tool builds on the `read_resource` feature and opens the channel side of the stack.


vabatta:

PR #785, opened 2026-08-13
feat(office-mcp): list a meeting's transcripts and their handles

vabatta:
## What this answers

Was this Teams meeting transcribed, and — for a recurring series — which occurrence.

`list_chats` already reports a `meeting_uri` on every meeting chat, and until now nothing took it:
the handle named a meeting rather than opening one. `list_meeting_transcripts` is the tool that
takes it. It resolves the join URL to a meeting, lists that meeting's transcripts, and returns one
handle per transcript with the times transcription began and ended. It returns none of the words
that were said, and there is no video anywhere in this connector.

Meeting discovery stays where it is. A meeting chat *is* the index — its `topic` is the meeting's
subject and its recency is what orders it — so "the pricing call last Tuesday" is identified by
`list_chats` and this tool takes the handle it minted. No `Calendars.Read`, no second way to ask
which meeting.

## The decisions worth knowing

**Five statuses, because there are five different things to do.** Four of them arrive from Graph as
the same bytes: an empty collection. `not_ready` means wait; `not_transcribed` means stop retrying;
`scan_incomplete` means nobody can tell; `meeting_not_found` means the join URL matched nothing.
Collapsing any pair of them produces a wrong answer a caller cannot detect — "there is no
transcript" about one that is ten minutes away, or "wait" for an occurrence that ended last month.

**The wait/stop inference reads the window, not the meeting.** Microsoft publishes neither a
processing status nor an availability SLA for transcripts, so this is an inference over one
deliberately generous allowance, and it says so in the description rather than pretending to be a
fact. Reading it off the meeting would be wrong in exactly the case the window exists for: a
recurring series is one meeting with one `endDateTime` that can sit years in the future, which would
make every empty window "still processing", including one bracketing an occurrence that was never
transcribed. Either the window's own end or the meeting's may settle it; absent evidence settles
nothing and answers `not_ready`, because one wasted call is cheaper than a wrong answer.

**`scan_incomplete` claims nothing and offers no remedy, and both halves are load-bearing.** Past
the scan cap the walk stops mid-collection, so an absence inside the window is an absence over a
prefix — which is no absence. Reporting `not_transcribed` there would say "retrying will not help"
about a collection nobody read to the end. And the natural-sounding remedy is a loop: the window is
ours, applied to the transcripts *after* Graph has answered, because Graph documents no filterable
date on this collection. A narrower window is the same bare request over the same transcripts with
the same answer. A test drives a wide window and a narrow one — the narrow one bracketing a
transcript that genuinely exists past the cap — and asserts one request and one answer for both. A
dead end stated plainly beats an actionable-sounding loop; the channel browser already paid for that
lesson with its circular reply advice.

**"Newest first" is a promise about the collection, not about a page of it.** Graph documents
`$select`, `$filter` and `$top` here and no `$orderby` at all, so the order it answers in is not a
contract. The window is therefore taken whole — to the cap — ordered, and only then cut to `limit`.
Cutting first returns an arbitrary handful sorted among themselves and calls them the latest, which
is a wrong answer with the shape of a right one. Past the cap the promise is honestly weaker, "the
newest of what was READ", and every description is worded to that bound rather than to the meeting.

**One claim was proved rather than carried.** `scan_incomplete` tells a model "this meeting has more
transcripts than one call reads" — a claim about the meeting, true only while a short walk can have
been stopped by nothing but a cap. Graph does answer empty pages that still advertise more, and the
SDK's own page walker reads one as the end of the collection. So a test pages a four-transcript
meeting as `[3 + nextLink]`, `[nothing + nextLink]`, `[the newest]` and asserts all four come back
with the scan reported complete. Without that, the tool would have told a caller a four-transcript
meeting was too large to read and lost the newest transcript in it.

**`include_scan_completeness` is opt-in, and one flag would have been two facts.** That the window
held more than `limit` a caller can already see — a full list may have more behind it, a short one
is all there was. That the read stopped at the cap it cannot, and the two have opposite remedies:
raise `limit` for the first, nothing at all for the second. Where nothing came back at all it is not
opt-in, because an empty answer is worthless without it: `status` says it.

**A third kind of 403.** Microsoft Graph access to Teams meeting transcripts is a tenant-wide Teams
setting that is off by default, and Graph reports it with the same status and the same outer code as
a missing permission. Both of the existing remedies are wrong for it — no permission is missing, and
re-consent changes nothing — so `GraphFailure` now carries `inner_code`, and the seam branches on it
to name a Teams administrator, the admin-centre path and the cmdlet, and to rule out re-consent
explicitly. Recognition is by inner code and never by message text, which Microsoft documents as
subject to change. In an untouched tenant this is the first answer this tool gives.

## What it adds to `shared/`, and why that is vocabulary

`shared/meetings.py` holds how a meeting is reached (the join-URL `$filter` and its escaping), which
occurrence was asked about (`OccurrenceWindow`), how far "newest first" is true
(`newest_in_window`, `MAX_ARTIFACT_SCAN`), and the delay allowance behind the wait/stop inference.
`shared/handles.py` gains the transcript handle family.

None of those is a fact about transcripts. They are facts about a *meeting* and about the artifact
collections Graph hangs off one — reached through the same resolve, ordered the same way (which is
to say not at all), answered empty in the same ambiguous way. That is the test for belonging in
`shared/`: a promise a second tool over the same meeting would otherwise make for itself, where the
disagreement is invisible from outside. A caller cannot see two tools disagreeing about "the latest
occurrence"; it can only see one of them being wrong.

This tool is currently the only caller, and the vocabulary still goes there. The alternative is not
"keep it local and move it when a second caller arrives" — it is a later diff that moves a promise,
reviewed as a refactor rather than as a decision, with the window spending that whole time as one
tool's private helper. A private helper is exactly what the next caller copies instead of importing.
The permission names live there for the same reason: layering rule 4 forbids one tool importing
another, so a permission spelled in a tool file is a permission the next module touching that
resource has to re-type — and a permission spelled twice is a scope Entra can reject, which fails
sign-in for every user.

The handle is a family of its own, `teams:///transcripts/{meetingId}/{transcriptId}`, rather than a
path nested under the meeting family: nesting would make the segment after `meetings/` a join URL in
one shape and a meeting id in the other, and no parser can tell those apart. It carries the resolved
meeting id rather than the join URL, so whatever reads a transcript does not repeat the resolve,
spend a second request and a second permission on it, and answer a 403 that could be about either.

`GraphCollection` becomes part of `graph_client`'s published surface, because the newest-first walk
takes a collection response the caller already awaited and the type has to cross the package's front
door rather than be reached past it.

## Layering rules

No new rule becomes assertable at this level. Rule 7 — nothing may address a single meeting
recording — still needs a recordings surface to protect, and its anti-vacuity guard would still fail,
so it stays absent and named where it is missing.

What does change is rule 6's reach: `shared/handles.py` now owns four handle families rather than
three, and the guard that pins which families it spells is updated with it — so a second speller of
the transcript grammar fails the build the same way a second speller of the chat grammar does. The
pagination bound is now asserted over two tools that pass different scan caps, which is the property
that number has to have: an empty page spends no scan budget, so the bound on a run of them is its
own and must not vary with either tool's cap.

### DIFF
--- a/services/office-mcp/README.md
+++ b/services/office-mcp/README.md
@@ -2,13 +2,14 @@
 
 An MCP server for Microsoft 365 via Microsoft Graph API.
 
-Users sign in with their own Microsoft account and the server acts as them. It exposes seven MCP
+Users sign in with their own Microsoft account and the server acts as them. It exposes eight MCP
 tools so far — `get_me`, the signed-in user's own profile; `list_chats`, their Microsoft Teams chats
 most recently active first; `list_teams`, the teams they are a member of; `list_channels`, the
 channels of one of those teams; `browse_channel`, what was posted in one of those channels;
-`search_messages`, full-text search across every Teams message they can see; and `read_message`,
-one of those messages in full — each one a file of its own, and more land in later PRs, stacked on
-top of this one, one tool per PR.
+`search_messages`, full-text search across every Teams message they can see; `read_message`,
+one of those messages in full; and `list_meeting_transcripts`, whether a Teams meeting was
+transcribed and a handle for each transcript — each one a file of its own, and more land in later
+PRs, stacked on top of this one, one tool per PR.
 
 ## Layout
 
@@ -33,7 +34,9 @@ Entra at startup. A forgotten permission means sign-in fails. `tests/test_app.py
 
 **`shared/` is the cost of file-per-tool architecture.** It holds the things two files must agree on:
 `handles.py` (the `teams:///` grammar), `messages.py` (Teams message shape, sender normalization,
-HTML unwinding), `identity.py` (signed-in user profile), `seam.py` (token exchange, error advice).
+HTML unwinding), `meetings.py` (join URL to meeting, which occurrence a window means, how far
+"newest first" holds), `identity.py` (signed-in user profile), `seam.py` (token exchange, error
+advice).
 Use `shared/` when two tools would otherwise need identical copies and differing copies would be
 visible bugs: handles one tool mints and another rejects; two answers to "who am I"; refusals that
 sound inconsistent. Don't put tool-specific things there: descriptions, parameters, output, requests.
@@ -81,6 +84,8 @@ Tools redeem permissions per call via On-Behalf-Of. No permission = no consent =
 | `Team.ReadBasic.All` | Delegated | No | `list_teams` |
 | `Channel.ReadBasic.All` | Delegated | No | `list_channels` |
 | `ChannelMessage.Read.All` | Delegated | Yes, in most tenants | `browse_channel`, `search_messages`, `read_message` (channels) |
+| `OnlineMeetings.Read` | Delegated | No | `list_meeting_transcripts` (resolving a join URL to a meeting) |
+| `OnlineMeetingTranscript.Read.All` | Delegated | **Yes** | `list_meeting_transcripts` |
 
 Multiple tools naming the same permission is normal. It is not duplication to remove. Each tool
 declares its own tuple because that tuple words its own 403 and AADSTS65001 messages.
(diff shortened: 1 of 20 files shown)


vabatta:

PR #786, opened 2026-08-13
feat(office-mcp): list meeting recordings and who may access them

vabatta:
# `list_meeting_recordings` — whether a call was recorded, and who may get at it

The tenth tool, and the second half of the meeting question. `list_meeting_transcripts` says what
was *said*; this says whether there is a recording at all, how long it runs, and whether Microsoft
would let the signed-in user download it. It takes the same `meeting_uri` `list_chats` reports on a
meeting chat, scopes a recurring series with the same `started_after`/`started_before`, and answers
in the same five-status shape — a model that learned one lister has learned this one.

## A recording is answered as metadata and availability. The bytes are never returned, by anything

This is the decision the whole tool is built around, so it is worth stating plainly rather than
discovering in a field list. Graph serves `…/recordings/{id}/content` as one inline `video/mp4`
stream with no ranged-fetch contract documented on that path; a Teams meeting can run thirty hours;
and a model cannot watch video. An MP4 encoded into a tool result is hundreds of megabytes of text
that answers nothing. `recordingContentUrl` is no better — it is a Graph URL that opens only with
this connector's own bearer token, so handing it to a caller is either useless or a token leak.
(Delegated download is organiser-only anyway, so for most callers it would also just fail.)

What comes back instead is what a caller can act on: that a recording exists, when it started and
stopped, a derived `duration_seconds`, who may download it, and `content_correlation_id` —
Microsoft's own identifier linking a recording to the transcript of the same call, which is the
route from "there is a recording nobody may download" to "here is what was said in it".

**Layering rule 7 is the ratchet on that**, and it arrives with this PR because this is the tool it
protects. No module may address a single `callRecording`: the two SDK names that do
(`by_call_recording_id`, `recording_content_url`) are forbidden anywhere under `src/`. The rule was
written into `tests/test_layering.py`'s numbering from the start and named there as absent, under
the same discipline every other rule arrived under — a rule may not land before its guard would
fail if the rule stopped forbidding anything. It lands as one class and changes nothing else.

Its guard finds the recordings listing by what it *does*
(`online_meetings.by_online_meeting_id(...).recordings`) rather than by where it lives, because the
rule is permanent and one tool file's home need not be: a guard anchored on a path fails the day the
listing moves and then reads as "delete me" when what it meant was "it moved". Matching is on
attribute names through the AST, so the module that has to explain in words why Microsoft's
`recordingContentUrl` is not passed on can keep saying it.

## Two artifacts, two tools — argued rather than assumed

One tool listing a meeting's transcripts *and* its recordings is the tidier surface, and it loses
answers. Microsoft gates the two independently, and one gate is shut by default: Graph access to
transcripts is a tenant-wide Teams switch that is off until an administrator turns it on, and while
it is off every transcript call answers `403` with no request-side workaround — whereas that
control applies to transcript resources only, and neither recordings reference page publishes a
tenant control or an inner error code of its own. The two delegated permissions are separate and
separately admin-consented, so a tenant can grant either without the other.

A combined tool in a default tenant therefore has two options and both are bad: fail the whole call
and say nothing about a recording that was perfectly reachable, or carry a verdict per artifact —
two statuses, two refusals-as-data, and the "read `status` first" shape that makes either answer
actionable replaced with something a model has to unpick. It would also blunt the one thing a 403 is
good for here, since every tool in this connector names only the permissions its own request was
made under, which is how an administrator is sent after the permission that was actually missing.

The tests say this rather than the prose alone: one meeting, the transcript listing 403ing on the
tenant switch and the recording listing answering `available` in the same test, over the protocol.

The standing cost is that the two tool files now share a lot of near-identical wording, and rule 4
forbids them importing it from each other. What holds them in step is tests rather than an import —
the shared status vocabulary and the shared answer shape are asserted across both, and each tool's
own distinguishing sentence is pinned per tool, so an edit that improves one wording and forgets the
other fails rather than ships.

## What this adds to `shared/`, and why it is vocabulary rather than a layer

Nothing new lands in `shared/`; two things already there stop being single-caller.

`shared/meetings.py` gains one name, `RECORDING_PERMISSION`, spelled beside the transcript one. It
lives there for the reason the transcript one does: it is the permission a *resource* costs rather
than the permission one tool's request costs, and it is named by two places that rule 4 forbids
importing each other. The resolve, the occurrence window, `newest_in_window` and the
`OccurrenceWindow.settled` inference are taken as-is — this PR is what makes them promises two tools
make about one meeting, and a caller cannot see two tools disagreeing about "the latest occurrence",
it can only see one of them being wrong.

`shared/identity.py` becomes load-bearing. Microsoft's organiser-only download rule is worth
*reporting* rather than reciting: a constant string on every recording says nothing, and the
organiser's identity alone does not answer the question either, because Graph returns
`meetingOrganizer.user.displayName` as null in every documented sample — so a model would be handed
a bare object id to compare against something it has not fetched. The comparison is therefore made
here, which costs one `GET /me` under `User.Read`, and `content_access` says which side of the rule
the signed-in user is on. That call is `get_me`'s question, so it is asked in `get_me`'s module: two
`GET /me` calls under two projections would be two answers to the one fact every other answer on
this connector is correlated against. It is spent only when there is a recording to say it about, so
the common "nothing was recorded" answer still costs two requests.

That is what makes both of these vocabulary and not a layer: each is a thing two tools would
otherwise each need a copy of, where the difference between the copies would be a bug a caller could
see. Nothing that one tool could own moved in — the description, the arguments, the answer shape,
the requests and every refusal only this tool can explain are all in the tool file.

## The organiser-only rule, and the wrong answer it exists to prevent

> In delegated permission scenarios, getting callRecording content is supported only for the meeting
> organizer. Meeting participants don't have permission to download meeting recordings. … For online
> meetings, tenant admins can unblock meeting participants to download meeting recordings.

The asymmetry is the point: the *metadata* is not so restricted — the same page says it is available
to users on the meeting invite. So for most meetings a participant asks about, the recording is
visible and its video is out of reach, and answering "there is no recording" would be a wrong answer
nobody could detect. Existence and reachability are therefore separate fields, an unreachable
recording is always listed, and both the tool description and the result schema say in as many words
that `organizer_only` must never be reported as a missing recording.

`content_access` is deliberately not a promise about a download: an administrator can block
recording downloads tenant-wide, and can equally have unblocked participants. It says what
Microsoft's documented default makes of *this* user, which is the most a caller can be told without
guessing.

## One inherited sentence that turned out to be false

Every list-shaped tool here says "that is all of it" by coming back short of `limit`, and everywhere
else that is honest because those walks follow Microsoft's paging to the end of the collection. The
meeting listers do not: they read to the artifact scan cap and apply the occurrence window to what
came back. So a series recorded daily for most of a year answers with two recordings against a
`limit` of twenty while recordings of that same window sit among the ones never read — a short list
meaning the exact opposite of what the convention says it means, on the one meeting shape where a
caller has no way to notice.

The new tool never shipped that sentence, and the transcript lister's copy of it is fixed here with
a test that runs the case rather than reading the words. No new field: `scan_incomplete` already
separates the two cases and is already opt-in, so what changed is the claim — fewer than `limit`
means the window holds no more *than was read*, which is the whole window for any meeting inside the
cap and for no meeting past it.

## Also in this PR

- The transcript lister's description told a model that nothing on this server answers whether a
  meeting was recorded. That stops being true here, and a description is protocol surface a model
  reads as fact — so it now names `list_meeting_recordings` and says why a refusal on one says
  nothing about the other.
- `OnlineMeetingRecording.Read.All` joins `REQUESTABLE_PERMISSIONS`, the hand-written list that is
  the only thing able to catch a misspelled permission — every other check compares the tool files
  against a union derived from those same files, so a typo would be on both sides of the comparison
  and would fail every sign-in for every user.
- README: the permission table's new row and the paragraph it needs, plus the asymmetry an operator
  has to know — a deployment where `list_meeting_transcripts` 403s and `list_meeting_recordings`
  answers is a correct deployment in an untouched tenant. The chart values say the same thing where
  an operator reads them.

## Verification

`ruff check`, `ruff format --check`, `basedpyright` and `pytest` are green. In-process, the app
advertises exactly the ten tools, and `app.GRAPH_SCOPES` is the registry's own tuple — derived from
the tool modules in declaration order, carrying every permission those tools declare and nothing
else.

## The curated presets

This is the top of the stack, so it is the first commit at which every tool a use-case preset names actually exists — and therefore the first at which one can be offered. `TOOLS_PRESET` gains six named surfaces beside `teams`, one per use case, and the README gains the table an operator reads before asking their administrator for anything:

| preset | what it can do | tools besides `get_me` | permissions | admin consents |
| --- | --- | --- | --- | :-: |
| `teams-chat` | name the live conversations — not read them | `list_chats` | `User.Read`, `Chat.Read` | 0 |
| `teams-messages` | find a message anywhere and read it in full | `list_chats`, `search_messages`, `read_message` | + `ChannelMessage.Read.All` | 1 |
| `teams-channels` | walk a team's channels and read what was posted | `list_teams`, `list_channels`, `browse_channel` | `User.Read`, `Team.ReadBasic.All`, `Channel.ReadBasic.All`, `ChannelMessage.Read.All` | 1 |
| `teams-transcripts` | find a meeting and read what was said | `list_chats`, `list_meeting_transcripts`, `read_transcript` | `User.Read`, `Chat.Read`, `OnlineMeetings.Read`, `OnlineMeetingTranscript.Read.All` | 1 |
| `teams-recordings` | say whether a meeting was recorded and who may get at it | `list_chats`, `list_meeting_recordings` | `User.Read`, `Chat.Read`, `OnlineMeetings.Read`, `OnlineMeetingRecording.Read.All` | 1 |
| `teams-meetings` | both of the above for one meeting | `list_chats`, `list_meeting_transcripts`, `read_transcript`, `list_meeting_recordings` | + both meeting permissions | 2 |
| `teams` | everything | every tool there is | all eight | 3 |

The table names the tools and says what each surface can *do*, because a count alone misleads: `teams-chat` is the only preset asking for **no** administrator, and the reason it costs nothing is exactly that it cannot read a message — the two tools that can both declare `ChannelMessage.Read.All`. `get_me` is always on, so every row is one tool wider than its third column. `teams` stays *derived* — every tool in the registry — so it needs no maintenance as tools land; the rest are named sets. The names carry a product axis from the start, so `outlook-*` and `sharepoint-*` join the table as those tools land without re-cutting these.

**The `teams-transcripts` row is what the knob was built for.** Reading meeting transcripts costs **one** admin consent and does not drag in `ChannelMessage.Read.All`, the permission to read every channel message in the tenant. `list_meeting_recordings` is *not* behind the tenant-wide Teams transcript switch, which is why recordings and transcripts stay separately selectable: a tenant that has never turned that switch on can still have recordings.

### What the tests hold

A preset that drifted would still resolve, still register and still start. The only visible difference would be a permission on a consent screen a tenant already agreed to — and that cannot be taken back without every user signing in again. So three things are asserted:

- **The table above is written out and checked per preset**, not derived from the mapping, because a derivation agrees with any mistake in it. This is the one place the promise "one admin consent and no `ChannelMessage.Read.All`" is checkable at all.
- **Every *argument* a tool requires has a producer inside the same preset.** Per argument rather than per tool, because `browse_channel` takes a `team_id` and a `channel_id` from two different tools and a check keyed by tool passes while one of them is unreachable.
- **The arguments are read off the live schemas — the whole schema, not just its top level.** `search_messages` states its nine criteria as a `required` inside each branch of an `anyOf`, which a top-level read misses entirely. Each required argument must be classified as minted by another tool or composed by the caller; nothing may be neither.
- **The classification itself is checked rather than asserted by comment.** An argument that names a tool in its own description must be recorded as minted — which is what stops a handle being filed as the caller's to compose, silently exempting its tool from the reachability check.

Why the reachability check needs to exist at all: `teams-messages` without `search_messages` asks for the **identical** three permissions — `User.Read`, `Chat.Read`, `ChannelMessage.Read.All`, because `read_message` declares the last two itself — and strands a `read_message` nothing in the preset can address. Permission review cannot catch that one. It stays a check on the presets *we ship* rather than a mechanism in the registry: a hand-written `TOOLS_ENABLED` may still name a consumer without its producer, and the tool's own refusal names the tool the handle comes from on first use.
- **No curated preset may resolve to the whole surface**, so one that quietly widened would fail rather than read as a narrow deployment while asking for everything.

`NEEDS_ADMIN_CONSENT` can finally be pinned in both directions. It was written complete while most of its permissions were still ahead of the tools that declare them, so a misspelled entry sat unchecked; every permission this connector asks for now exists, so a name in the table no tool can declare is a failing test rather than an assertion from the manifest the day its tool lands.

The seam all of this rests on — the two variables, `resolve`, the manifest, the description scan — is in `office-mcp/feat/first-tools` (#779), where `register_tools` and the scope list are born.


vabatta:

PR #787, opened 2026-08-13
fix(office-mcp): make "newest first" true of the collection, and never assert an absence the scan did not see

vabatta:
Two commits fix two defects in the two meeting listers, and correct two statements that a model
reads. Commit `e8be42e2` makes the two fixes. Commit `e4955b25` corrects the two statements.
Together they touch 12 files, with 1123 insertions and 109 deletions.

## The two fixes

The transcript lister and the recordings lister repeated the same two statements: the
`collect_pages` call and the `sorted` call. Those two statements now live in one function,
`features/transcripts.newest_in_window`, and both listers call it.

`began_at` is private again as `_began_at`. `recordings` imports the shared function instead of a
sort key.

### A — the sort ran after the cut

`collect_pages(..., limit=limit)` stopped the read at `limit` items in Graph server order.
`sorted(..., reverse=True)` then ran over those items only. A model that asked for the 3 newest
recordings got 3 arbitrary ones, sorted among themselves. Nothing in the answer said so.

`list_meeting_transcripts` had the same defect. `newest_in_window` keeps every artifact that the
occurrence window holds. It sorts that result. It cuts the result to `limit` last.

**What one call costs now.** Before the fix each lister called
`collect_pages(first_page, client, limit=limit, matches=window.holds)`, so the read stopped at
`limit` matched artifacts. `newest_in_window` passes `limit=MAX_ARTIFACT_SCAN` and
`max_scanned=MAX_ARTIFACT_SCAN` (`features/transcripts.py:768`). Every call therefore reads the
meeting's artifact collection, up to 200 artifacts, in as many pages as Graph chooses. The common
case costs more, and a sort requires that cost. Microsoft documents no `$orderby` on either
collection, so a tool cannot name the newest artifact before it reads the collection.

`MAX_ARTIFACT_SCAN = 200` caps that cost. Items are what `collect_pages` can cap. 200 is four times
50, and 50 is the largest `limit` that either tool accepts (`MAX_TRANSCRIPTS`, `MAX_RECORDINGS`).

**Where the cap lives.** `MAX_ARTIFACT_SCAN` is in `features/transcripts.py`.
`features/recordings.py` imports it, and `server/tools.py` interpolates
`transcripts.MAX_ARTIFACT_SCAN` into the recordings descriptions as well. One home is deliberate:
the cap is one promise about two artifacts, and two constants would drift apart.
`graph_client/pagination.py` names `features/transcripts.MAX_ARTIFACT_SCAN` as the tighter cap of
its second way to bound a read. `features/__init__.py` records the newest-first walk among what
`recordings` borrows from `transcripts`. `tests/test_layering.py` passes at the head with that
import in place.

The cap number is also in the words a model reads. Those places are both `limit` descriptions, both
list fields, both `truncated` fields, both `status` descriptions and both tool descriptions.

### B — one answer could report `not_transcribed` together with `truncated: true`

"A retry will not help" and "there is more" cannot both be true, and a caller cannot see which one
is wrong. The defect needed a meeting with more artifacts than the cap.

A tool can report an absence only from a collection that it read to the end. A read that
`MAX_ARTIFACT_SCAN` stopped therefore reports no absence. It reports **`scan_incomplete`**, in both
tools. `_absence(scan_stopped_short=collected.truncated, settled=...)` takes
the flag and the verdict from one value. Each lister reports `available` when the window holds
artifacts, so `truncated: true` can accompany only `available` or `scan_incomplete`.

`scan_incomplete` is a fifth value of the existing `status` field. It is not a new field, not a new
parameter and not a new tool.

**What an older consumer sees.** `status` is a plain `str` on both models (`transcripts.py:465`,
`recordings.py:188`). The fifth value is therefore not a schema-visible enum change. A consumer that
branches on the four old values falls through silently, so such a consumer needs a default branch
for the new word. Both tool descriptions name all five values.

## Two statements that a model reads, corrected in the second commit

Both listers apply the occurrence window on the client side, and the two Microsoft references are
the reason. `calltranscript-get` and `callrecording-get` both document `$select` as the optional
query parameter, and neither one names `$orderby`. Each reference shows one `$filter` example, and
each of those filters on `contentCorrelationId` (`calltranscript-get` Example 11,
`callrecording-get` Example 5). The request therefore goes out bare, and the read takes up to
`MAX_ARTIFACT_SCAN` artifacts in Graph's own order. A feature test per lister asserts that the
request URL carries no `$filter`, no `$orderby` and no `$top`.

### C — `scan_incomplete` offered a remedy that the mechanism cannot keep

The first commit told the caller to narrow `started_after`/`started_before` and to ask again. Each
lister applies the window to the artifacts after Graph answers. A narrower window therefore sends
the same request, reads the same artifacts and returns the same status.
`test_narrowing_the_window_cannot_reach_past_the_scan_cap` asserts one identical request URL for a
wide window and a narrow one.

The status now states that nothing can help. It tells the caller to stop and to report that the
answer is not determined. `test_the_unreadable_answer_tells_a_caller_to_stop_rather_than_to_retry`
asserts that the `status` description holds "There is nothing to try", "Stop, and report" and "do
not ask again". It also asserts that the old remedy sentence is gone.

### D — "newest first" was a promise about the collection

Past the cap, "newest first" describes the artifacts that the call read. Every place that makes the
promise now names which of the two it means, and each place names `truncated` as the flag for the
difference. The cap itself is unchanged: with no `$orderby`, a larger cap reads more artifacts and
still cannot promise the newest.

`truncated` has two causes now, and the descriptions separate them.

- The window holds more artifacts than `limit`. The entries are the newest of the window, and the
  caller can raise `limit` to 50.
- The meeting holds more artifacts than one call reads, which is `MAX_ARTIFACT_SCAN`. The order then
  covers the artifacts that the call read, so the first entry can be older than the meeting's newest
  artifact. No argument to the tool reads further, because each lister applies the window after
  Graph answers.

Which cause happened is not always visible. Both `truncated` descriptions tell a model to call the
first entry the most recent of what was read. Fewer entries than the caller asked for, together with
`truncated: true`, is always the second cause.

## What the branch verified

- **The journey tests, in-process over the real MCP protocol** (`fastmcp.Client` against the server
  that `create_app` builds, and Graph through respx). Each one passes every id and handle from one
  answer into the next call. `test_a_model_walks_from_its_teams_to_a_reply_it_can_read` runs
  `list_teams → list_channels → browse_channel → read_message` on a reply handle that
  `browse_channel` mints. `test_a_model_walks_from_a_meeting_chat_to_what_was_said_in_the_meeting`
  runs `list_chats → list_meeting_transcripts → read_transcript`.
  `test_a_model_walks_from_a_meeting_chat_to_whether_the_call_was_recorded` runs
  `list_chats → list_meeting_recordings`.
  `test_a_handle_from_a_search_result_reads_the_message_behind_it` runs
  `search_messages → read_message`.
- **A protocol-level test for fix A** over both listers
  (`test_asking_for_the_latest_of_a_series_answers_with_the_latest`, parametrized over `transcripts`
  and `recordings`). The mocked collection answers a three-occurrence series oldest first, the call
  passes `limit: 1`, and the test asserts one entry `week-3`, `status: available` and
  `truncated: true`.
- **The first commit added tests for fix A**. The fix-A test runs over both listers. Six more tests
  are feature-level, three per lister. One test is
  `test_both_meeting_listers_teach_the_same_status_vocabulary`. The commit also replaced the two
  "four answers" description tests with "five answers" tests.
- **The shared vocabulary test** asserts four shared words:
  `{\"available\", \"not_ready\", \"scan_incomplete\", \"meeting_not_found\"}`. It reads them from the
  `status` field description of each output schema. It also asserts that `list_meeting_transcripts`
  names `not_transcribed`, that `list_meeting_recordings` names `not_recorded`, and that neither
  tool names the other word.
- **The second commit added tests for fixes C and D.** Three protocol-level functions run parametrized over both
  listers: `test_neither_lister_offers_a_remedy_its_mechanism_cannot_keep`,
  `test_neither_lister_promises_the_newest_past_its_scan_cap` and
  `test_a_meeting_larger_than_one_call_reads_answers_within_what_it_read`.
  Eight feature-level tests are also added, four per lister.
- **What the largest of those tests builds.** `_daily_series` builds `MAX_ARTIFACT_SCAN + 60`
  artifacts, which is 260. `limit: 1` answers `day-199`, and `truncated` is true. A wide window and a
  narrow window over the part that the call did not read both answer `scan_incomplete`. Both answers
  carry the same `truncated` value and the same empty list.
- **All CI gates on this branch pass.** This includes pytest, ruff check, ruff format --check, and
  basedpyright. `tests/test_layering.py` passes at the head.
- **The real process**, at the default port 9544 (`config.py` has `port: int = Field(default=9544,
  …)` and `.env.example` has `PORT=9544`). `/health` 200, `/probe` 200, and `/metrics` 200 in the
  Prometheus format. `/.well-known/oauth-authorization-server` and
  `/.well-known/oauth-protected-resource/mcp` both answer 200 and agree: `issuer` is
  `http://localhost:9544/`, `authorization_servers` holds that same value, and both documents name
  the `access_as_user` scope. An unauthenticated `POST /mcp` answers `401` with
  `WWW-Authenticate: Bearer resource_metadata=\"http://localhost:9544/.well-known/oauth-protected-resource/mcp\"`.
- **The readiness probe, both ways.** With Postgres reachable on port 9001, `/ready` answers 200 and
  `{\"status\":\"healthy\",\"checks\":{\"database\":true}}`. With an unreachable database port, `/ready`
  answers 503 and `{\"status\":\"unhealthy\",\"checks\":{\"database\":false}}`. `/health` still answers 200,
  and the service logs `ready.database_unreachable` as a structured event at warning level. It does
  not crash.

## Drift found, and how the branch resolved it

| Suspect | Finding |
| --- | --- |
| `collect_pages`'s `matches` / `max_scanned` have no production caller | **The suspicion holds for `max_scanned` only.** On the base branch `matches` had two callers and `max_scanned` had none. At the head each one has exactly one caller: `newest_in_window` passes both (`features/transcripts.py:768`). The meeting read passes its own tighter cap instead of `MAX_SCANNED_ITEMS = 1000`. A new docstring in `tests/graph_client/test_pagination.py` records that both parameters are used in production. The branch deleted nothing. |
| Item budget (teams/channels) against request budget (browse_channel) | **`graph_client/pagination.py` now records three ways to bound a read, in one place**, and the README points at that module. An inventory walks until it has `limit` items. A meeting's artifacts are walked to a tighter scan cap, because this code filters them and also sorts them. A channel's messages get no paged read, at one request per second per app per tenant. |
| The outcome vocabulary across two tools | **The fifth word is identical in both tools**, and only the settled absence differs (`not_transcribed` against `not_recorded`). `test_both_meeting_listers_teach_the_same_status_vocabulary` asserts the four shared words, and it asserts that neither tool names the other tool's absence word. |
| Ten descriptions, one voice | **One description was incomplete.** `get_me` now names where its `user_id` matches: `search_messages` matches `mentions` on it, and a recording reports `organizer_user_id`. It also names `email` as the value that matches a chat member from `list_chats`, because that list carries no ids. |
| Layering: no rule skipped or weakened | **`tests/test_layering.py` is unchanged by this branch.** All tests pass at the head. |
| The README | **Three passages were stale, and the branch corrected all three.** The module list named neither `transcripts` nor `recordings`. The status paragraph said four statuses, and it split the two absences on the meeting's end time. It says five now, the end of the window splits them, and the meeting is the fallback. The paragraph about bounded reads gave one way to bound a read, and it gives three now. |

## Deployability

- The CI job runs `helm dependency update` and then `helm template .` with
  `--values ci-values.yaml` where that file exists, and it exists for this chart. The job pins helm
  v3.19.0 (`azure/setup-helm` v4.3.1 with `version: v3.19.0`). This session could not repeat the
  render, because helm is not on its PATH.
- `scripts/render-values-schema.sh` **generates** `values.schema.json` from the base chart's schema
  and `values.additional.schema.json`, and nobody edits it by hand. This session could not run that
  script either, because it needs helm. The branch changed no chart value and no schema file: the
  one chart change is comment text in `values.yaml`.
- The branch adds **no configuration**. `config.py` and `.env.example` are not in the diff. Every
  Graph permission is a code constant, and there are eight distinct permission strings across
  `features/`.
- **What `values.yaml` records.** Eight delegated permissions. Two of them need admin consent
  independently of each other: `OnlineMeetingTranscript.Read.All` and
  `OnlineMeetingRecording.Read.All`. The last item is the tenant-wide Teams setting for Graph
  access to meeting transcripts, which is off by default. Until a Teams administrator turns that
  setting on, `list_meeting_transcripts` answers 403 with that remedy, and every other tool answers
  as before.

## Not changed, on purpose

- **`MAX_SCANNED_ITEMS = 1000`** stays the default in `graph_client/pagination.py` for the
  inventories. Nothing filters them, so the items kept are the work done, and that cap never binds.
- **`$top` on the two artifact collections.** Neither Microsoft reference documents it, and the two
  requests go out without it. A feature test per lister pins that.

## For the owner (the branch did not build these)

- **No cursor on the two meeting listers.** They take `meeting_uri`, `started_after`,
  `started_before` and `limit`. A caller cannot reach the artifacts past `MAX_ARTIFACT_SCAN`, and a
  narrower window does not help, because each lister applies the window after Graph answers. A
  cursor is the feature that would remove that limit.
- **No metric on the cap.** `scan_incomplete` needs a meeting with more artifacts than the cap, so
  it is hard to observe in production. A metric on the cap is the cheap way to see it. The branch
  added none, and `metrics.py` is not in the diff.

### DIFF
--- a/services/office-mcp/README.md
+++ b/services/office-mcp/README.md
@@ -16,7 +16,8 @@ src/office_mcp/
   auth.py                Entra auth: which app registration, and where its state lives
   logging.py metrics.py  cross-cutting, used by both sides below
   graph_client/          the Microsoft Graph transport — the official SDK, one caller's token
-  features/              what the connector does — one module per slice (identity, chats, channels, messages)
+  features/              what the connector does — one module per slice (identity, chats, channels,
+                         message search/read, meeting transcripts, meeting recordings)
   server/                how it's exposed over MCP — the tools, the errors they report, /ready
 ```
 
@@ -171,9 +172,14 @@ gets that string onto the wire.
   Wrap Graph work in `with graph_errors():`. Anything else stays the base `GraphFailure`.
 - **Paging follows `@odata.nextLink`** via `collect_pages`, replaying the URL verbatim, with both an
   item cap and a *scan* cap — the teams-mcp lesson that a filtered collection can walk a long way
-  for very few kept items — and a `truncated` flag so a partial answer never looks complete. Search
-  is not paged this way: `POST /search/query` takes stateless `from`/`size` offsets, so a search
-  tool resumes by re-issuing rather than by carrying a cursor.
+  for very few kept items — and a `truncated` flag so a partial answer never looks complete. There
+  are three ways a read is bounded here and `graph_client/pagination.py` sets all three out
+  together: an inventory walks until it has `limit` items, a meeting's artifacts are walked to a
+  tighter *scan* cap (they are filtered and ordered here, so the walk cannot stop at `limit`), and
+  a channel's messages are not walked at all — Graph allows about one request a second on a given
+  channel for the whole app, so `browse_channel` makes exactly one and `$top` is its window. Search
+  is not paged this way either: `POST /search/query` takes stateless `from`/`size` offsets, so a
+  search tool resumes by re-issuing rather than by carrying a cursor.
 - **The caller's token goes to `graph.microsoft.com` and nowhere else.** The SDK's bearer provider
   does not check its own allowed-hosts validator, so a redirect or an off-Graph `nextLink` would
   otherwise be handed a user's delegated credential.
@@ -330,13 +336,28 @@ Decisions worth knowing:
   over the protocol.
 - **`200 OK` with an empty `value` is an answer, not an error.** The `JoinWebUrl` filter never 404s,
   so "no match" is `status: meeting_not_found` — reported as not proof the meeting is gone.
-- **Four statuses, because there are four different things to do.** `available`, `not_ready`,
-  `not_transcribed`, `meeting_not_found`. `not_ready` and `not_transcribed` are the *same* empty
-  collection from Graph and the opposite advice — wait, or stop — and Graph publishes no
-  "processing" status and no availability SLA, so the split is inferred from the meeting's end time
-  with a deliberately generous allowance and the field says so. A meeting Graph gave no end time for
-  counts as `not_ready`: one wasted call is cheaper than telling a caller a transcript will never
-  exist ten minutes before it arrives.
+- **Five statuses, because there are five different things to do.** `available`, `not_ready`,
+  `not_transcribed` (`not_recorded` for the other artifact), `scan_incomplete`,
+  `meeting_not_found`. `not_ready` and the settled absence are the *same* empty collection from
+  Graph and the opposite advice — wait, or stop — and Graph publishes no "processing" status and no
+  availability SLA, so the split is inferred from the end of the *window that was asked about*,
+  falling back to the meeting's own end time where no window was, with a deliberately generous
+  allowance; the field says so. A meeting Graph gave no end time for counts as `not_ready`: one
+  wasted call is cheaper than telling a caller a transcript will never exist ten minutes before it
+  arrives. `scan_incomplete` is the fifth because an absence is only knowable from a collection
+  that was read to the end: a meeting with more artifacts than one call looks through, none of them
+  in the window, used to answer "there is none" *and* `truncated: true` in the same breath, which
+  cannot both be true and which a caller cannot see through. Now the scan that stopped short is the
+  answer, and it claims nothing.
+- **"Newest first" is a promise about the collection, not about a page of it.** Both meeting
+  listers take the whole window (up to a 200-artifact scan cap, which is this call's whole cost),
+  order it, and only then cut it to `limit` — because Graph documents no `$orderby` on either
+  collection, so a `limit` applied to the order Graph chose returns an arbitrary handful sorted
+  among themselves and calls them the latest. "The latest transcript of this series" is the
+  question the tools exist for, and that shape answered it wrongly with the shape of a right
+  answer. The order, the cut, the scan cap and the `truncated` flag are one function
+  (`features/transcripts.newest_in_window`) shared by both, which is why the recordings lister
+  inherited the bug and why both were fixed at once.
 - **The tenant switch gets its own remedy, keyed on the inner error code.** `403` +
   `GraphAccessToTranscriptsDisabled` is not a permission problem and not a consent problem, so the
   message names a Teams administrator and the cmdlet, and explicitly rules out re-consent and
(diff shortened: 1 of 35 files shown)


vabatta:

PR #791, opened 2026-08-14
fix(office-mcp): walk past an empty page, and take "there is more" off the tools

vabatta:
Two jobs on the same code: make an incomplete read impossible to misattribute, and take the `truncated` field off the tools.

## Job 1 — the shared walk

**What the installed `PageIterator` actually does.** Read from `.venv/lib/python3.12/site-packages/msgraph_core/tasks/page_iterator.py` (msgraph-core 1.5.1, matching `uv.lock`). `enumerate()` opens with

```python
page_items = self.current_page.value
if not page_items:
    return False
```

and `iterate()` treats a `False` return as the end of the walk (`if not keep_iterating: return`). It never looks at `odata_next_link` in that path. So **an empty intermediate page carrying an `@odata.nextLink` ends the SDK's walk** — cause (b) is confirmed from source, not inferred. Worse, the walk then stopped with `pause_index == 0` and a next link on the current page, which is exactly what the old `_stopped_short` read as "a cap stopped me": a four-transcript meeting paged `[3 + nextLink]`, `[empty + nextLink]`, `[newest]` was told it "has more transcripts than one call reads (200)", and the newest transcript was never returned at all.

**The fix.** `collect_pages` now drives the pages itself instead of calling `iterate`, keeping the two parts of the SDK's walker worth having (`enumerate` reads the current page from wherever the last stop left it; `next` replays the cursor and deserializes the answer as `type(first_page)`). The loop's own stop conditions are: a cap, the end of the collection, or the request budget. An empty page carrying a next link is followed. `enumerate`'s return value is ignored, because it conflates "the page ran out" with "the callback stopped me" — the callback's own flag is what is read.

**The request bound, stated where a caller can see it** (`graph_client/pagination.py`, which is the one place the connector's three bounding shapes are set out together, and the README's Graph-client section). Following empty pages needs a bound or a pathological collection spins, and `MAX_ARTIFACT_SCAN` bounds artifacts rather than requests — the residual the module docstring used to admit. The budget is `max_scanned + MAX_EMPTY_PAGES + 1` requests, the `+1` being the page the caller already fetched. It is *derived* rather than picked, and that is the point: every page carrying at least one item spends scan budget, so at most `max_scanned` requests can go on pages that carried anything, and the fixed allowance (10) bounds the rest. A collection answering with items therefore cannot reach the budget, so it can only ever bind on one that answers with empty pages.

Exhausting it **raises** rather than returning what it has. An answer cut short for that reason would be byte-identical to one cut short by a cap and would mean something entirely different, and the honesty every tool above is built on is that a short answer means a cap. That keeps the property Job 1 asked for exactly: **a returned answer that stopped short can only have hit a cap**, so `scan_incomplete`'s sentence about the meeting's size is simply true, with no fallback wording needed.

Pinned by `tests/graph_client/test_pagination.py`: `[3 + nextLink]`, `[empty + nextLink]`, `[newest]` returns all four items with `capped` false; an endless run of empty pages raises after exactly `max_scanned + MAX_EMPTY_PAGES + 1` requests. And at the feature level by `test_a_meeting_graph_pages_through_nothing_is_read_to_its_end`, which is the reproduction verbatim: a four-transcript meeting in that page shape now lists all four newest-first and reports `scan_incomplete: false`.

`CollectedItems.truncated` is renamed `capped`, because after this it means one thing — a cap stopped the walk with more on offer.

## Job 2 — `truncated`, per tool

| Tool | What happened | Why | What a caller loses, and how completeness is conveyed now |
| --- | --- | --- | --- |
| `list_chats` | **Removed** | Derivable, exactly | A window filled to `limit` may have more behind it; fewer than `limit` is every chat there is. That second half is only *exactly* true because of Job 1 — before it, a short page could be the empty-page artefact. Nothing is lost; the remedy (a wider `limit`, no cursor) is on `limit` and in the description. |
| `list_teams` | **Removed** | Same | Same. |
| `list_channels` | **Removed** | Same | Same. |
| `browse_channel` | **Removed** | Unnecessary | This tool is one request and one page by design, and its answer is now described as a window rather than flagged as a partial one. Genuinely lost: "Graph's page carried a next link" and "a thread was cut". Neither has a remedy the description does not already give (raise `limit`, or `search_messages` with `sent_before`; a thread's older replies are a dead end), and the second is derivable — Graph expands up to 200 replies before it pages a thread, so a thread it paged comes back full to the 10-reply window either way. The one real loss is that a page shortened by dropped system messages no longer says so per call, so the description now says it unconditionally and in capitals: fewer posts than `limit` is **not** evidence the channel holds no more. |
| `search_messages` | **Removed** | Derivable | `next_offset` is non-null exactly when the index holds more, and says where to continue. Nothing lost. |
| `read_transcript` | **Removed** | Derivable | Same: `next_offset` non-null ⟺ more turns. Nothing lost. |
| `list_meeting_transcripts` | **Opt-in parameter**: `include_scan_completeness` → `scan_incomplete` | Half derivable, half not | See below. |
| `list_meeting_recordings` | **Opt-in parameter**, identically | Same | See below. |

### The two artifact listers, and the collapse

`truncated` there carried two facts with opposite remedies: *the window held more than your `limit`* (raise `limit`) and *the read stopped at the 200-artifact cap* (nothing helps, and the first entry may not be the meeting's latest). Collapsing them forced the caveat belonging to the rare case onto every answer that filled its window. The two are now separated rather than preserved or merged:

- **The window held more than `limit`** is what a full `transcripts`/`recordings` list says, the same reading as every other window here, with the remedy on `limit`. `newest_in_window` no longer folds `len(newest) > limit` into the flag.
- **The read stopped at the cap** is not derivable in general — with a window in play, a short list says nothing about the scan — has no remedy, and is the only thing bounding "newest first". So it survives as an opt-in: `include_scan_completeness` (input, default `false`, so a client that does not want it never sees it) reports `scan_incomplete` (output, `null` unless asked). It decides the answer's shape and never the work: the same two requests either way.

Both promises the last two rounds bought stay exact. `transcripts`/`recordings` still says the order is over what was read and names the cap, and now points at the parameter for the one question that turns on it. And nothing got quieter where it mattered: **an empty answer over a scan that stopped short is still `status: scan_incomplete`, asked for or not**, because an absence asserted over a prefix is the wrong answer that started all of this — while the same fact about a *non-empty* answer is the rare caveat behind the opt-in.

## Tests and docs

No test was weakened or deleted; assertions were migrated where the field legitimately went away, and every behavioural guarantee kept its own assertion. `test_the_whole_surface_speaks_one_language` now asserts the opposite convention — no `truncated` on any tool, `next_offset` on the two that page — with the reason recorded, so a new tool re-introducing the word re-introduces the ambiguity and fails. The full-window cases keep their newest-first assertions and gain `scan_incomplete is False`, which is the collapse being resolved rather than described.

All four gates pass on this branch: `uv run pytest`, `uv run ruff check .`, `uv run ruff format --check .`, `uv run basedpyright .`


vabatta:

PR #792, opened 2026-08-14
feat(backstop-mcp,main): fetch custom fields on demand instead of glossary prefetch

nicolae536:
## Summary
- Stop appending custom-field glossaries on `tools/list`. `list_custom_fields` takes a list of types (`organizations`, `people`, `accounts`, `opportunities`, `products`, `party`) and returns the mapped definitions for those types, including layout fields and `selectOptions`.
- Map those names onto the live Backstop Beans (`OrganizationBean`, …, `PartyBean`). Lookups are case-insensitive. Omit `contacts` / `employees` — those are not valid `entityType` enum constants.
- Cache the full `/custom-field-definitions` walk in process memory (`BACKSTOP_CUSTOM_FIELD_SCHEMA_TTL_MINUTES`, default 60, max 24h; `page[limit]=1000`, no LOV fetch). Optional `refresh=true` only when the user reports a missing field. First caller fills the catalog for everyone.
- Remove glossary middleware (and the empty leftover package), the Postgres snapshot table, the service-account env pair, boot warmup, `/lov-entries`, and `BACKSTOP_CUSTOM_FIELD_OVERRIDES`.

## Test plan
- [ ] `list_custom_fields` with `entity_types=["organizations","people"]` returns only those groups; keys are tool names; items keep Bean `entityType`
- [ ] `contacts` / `employees` and an empty `entity_types` list are rejected by the schema
- [ ] First catalog `get()` paginates at 1000 with no `include` / no `/lov-entries`; a second `get()` within TTL does not call Backstop
- [ ] `refresh=true` fetches even when fresh; concurrent refresh/cold fills share one walk
- [ ] Failed fetch with a cache returns `cache: "stale"`; cold failure raises
- [ ] `/ready` checks Postgres only (no `custom_field_schema`); `get_person` / `get_organization` have no `glossary_meta`


nicolae536:
Closing so we can reopen this cleanup on the restored activity-history branch. GitHub would not retarget a stacked PR after #751/#772 were accidentally merged into #750 and then split back out as #793/#794.

PR #797, opened 2026-08-15
feat(office-mcp): list Teams chats and their meeting join URLs

vabatta:
Add the second MCP tool: `list_chats` returns Teams conversations ordered by recency. The tool enables three infrastructure pieces.

**Pagination fix.** `collect_pages` previously delegated page walking to `PageIterator.iterate`, which stopped on any empty page, including one with `@odata.nextLink` still set. Graph returns those, and the walk returned one item of four. Empty pages are now walked through rather than read as the end, as long as a next link is present. A run of more than 10 consecutive empty pages (`MAX_EMPTY_PAGES`) raises `GraphPagingUnending` rather than returning a truncated result, which distinguishes this failure from reaching `max_scanned`.

**Handle spelling.** `shared/handles.py` is added as the single module that builds or parses `teams:///` URIs. Currently it spells meeting URIs from a join URL. A handle is how one tool's answer becomes another tool's argument, and this works only while there is one definition. Placing the grammar in `shared/` now, before there are two tools, prevents two tools from disagreeing on the shape.

**Tool implementation.** `list_chats` calls `/me/chats` with `$expand=lastMessagePreview` to sort by actual recency (not `lastUpdatedDateTime`, which tracks membership and renaming, not message activity). Results order by `last_message_at`. Members list is returned only for topic-less chats, capped at 25 per Graph (reported as `members_may_be_incomplete` if full). Meeting join URL comes via the handle speller from `onlineMeetingInfo.joinWebUrl`. Uses `Chat.Read` permission (required for message preview).

**Layering tests.** `test_layering.py` now asserts two rules that were previously marked absent:
- Rule 4: No tool module imports another. The detector checks module references in import statements and package member access (e.g., `from .. import tools` followed by `tools.get_me.GRAPH_PERMISSIONS` uses the same file twice).
- Rule 6: One speller per handle family. The detector checks what a scheme literal is *used for* — concatenation, formatting, parsing, regex matching — not spacing. Literals in prose (descriptions, examples, docstrings) are exempt.

## Role in the stack

This PR adds `shared/handles.py`, the one module that spells or parses a `teams:///` URI, and mints the first handle family from it: `teams:///meetings/{joinWebUrl}`. `list_chats` is the only tool that produces one today, on a meeting chat's `meeting_uri`. It builds on `office-mcp/feat/first-tools` (#779), which added `get_me` and the seam (`shared/seam.py`, the `tools/` registry) this tool reuses.

Next: `search_messages` (#780) adds the chat and channel message handle families to the same module, reusing rule 6 (one speller per handle family) rather than minting a family of its own. Meeting discovery is not a separate tool — meeting chats arrive here with their join URL already in `onlineMeetingInfo`.


vabatta:

PR #798, opened 2026-08-15
feat(office-mcp): list the channels of one team

vabatta:
# `list_channels` — the channels of one team the signed-in user can see

`list_teams` names a team and stops there. This is the tool that gets inside one: given a
`team_id`, the channels in it — id, name, description, membership type, creation date. No message
content, and no new machinery. It is the second half of the channel inventory, and the first tool
on this connector to take an id another tool minted.

## What it answers, and why the pair is the unit

A channel id addresses nothing on its own. Every team has a `General`, so a channel *name* is
unique only inside its team, and Graph's own path for a channel message is
`/teams/{team_id}/channels/{channel_id}/messages/{id}` — team first, always. That is why this tool
takes a `team_id` rather than offering to find a channel by name, and why the answer's field
descriptions say the id addresses a channel only together with its team's. `read_message` already
takes a handle carrying both; this is where a model gets them from something other than a search
hit.

The other thing it answers is the naming question `list_teams` answers one level up.
`search_messages` reports a bare `channel_id` on every channel hit. Until now nothing could turn
that into "the release channel in Engineering". Now something can.

## Decisions worth knowing

**`$select` is a requirement here, not an optimisation.** Microsoft documents populating a
channel's `email` as "an expensive operation that results in slow performance", and there is no
switch for it other than selecting around it. So the field list is load-bearing and the test asserts
it against the query that actually goes on the wire — an unnoticed missing `$select` is a listing
that got slower for nobody's benefit, and nothing else would report it.

**No `$top` on `/teams/{id}/channels` either — documented unsupported.** The same reference gives
this collection `$select` and `$filter` and nothing else. `services/teams-mcp` shipped a `$top` here
and had to take it back out, so sending one buys a 400 rather than a narrower answer. `limit` is
therefore a window this connector applies while walking the pages Graph chose, exactly as
`list_teams` does, and `MAX_CHANNELS` bounds how many Graph requests one call may spend rather than
anything Graph would refuse.

**"Fewer than `limit` is all of them" is proved by test, not carried across.** That sentence is the
whole completeness signal on this surface — no flag, no cursor — and it is only true while the walk
follows Microsoft's paging to the end. Graph answers the odd page with nothing in it and an
`@odata.nextLink` still set, and the SDK's own walker reads that as the end of the collection.
Believing it would not merely drop a channel; it would turn a window with more behind it into "this
team has one channel", a claim about the user's own team that nothing checked. The test that says so
for *this* collection is what makes the sentence shippable. It was checked by breaking the walk: with
an empty page treated as the end, that test is the only one in the file that fails.

**The answer says what it cannot see.** Graph trims this collection to the caller — private and
shared channels they are not a member of are simply absent, not marked absent. So the description
says an absent channel is not evidence the team has no such channel, rather than leaving a model to
conclude it, and `membership_type` names which kind each visible channel is. A `membershipType` this
connector predates deserializes to `None` rather than raising, and a channel whose kind we cannot
name is still a channel that has to be listed.

**Two permissions for the channel inventory, on purpose.** `Channel.ReadBasic.All` is Microsoft's
least-privileged delegated scope over a channel's identity, which is all this collection returns,
and it needs no administrator. Reading what was *posted* in a channel is the broad
`ChannelMessage.Read.All` that most tenants gate behind admin consent. Keeping them separate is what
a tenant actually uses: refusing the broad one still leaves the inventory readable, and each tool's
403 names only the permission its own request needed rather than sending an administrator after one
that was never missing. That is asserted end to end, per tool, and not merely written in the README.

## What it adds to `shared/`

Nothing — and that is the decision, not an omission.

The tempting candidates are both non-vocabulary. The page-size ceiling is one: `list_channels` picks
the same number `list_teams` picked, over a different Graph collection, and two tools agreeing about
a number is not agreement about a *meaning*. Either is free to change it tomorrow without the other
being wrong, so a shared constant would create a coupling the domain does not have. The other is the
prose about paging a `$top`-less collection, which reads nearly the same in both files — and prose
that two tools share is prose that has stopped being about the tool it is on. Rule 4 exists for
exactly this, and the honest answer here is that this tool leans on `shared/seam.py` for its token
and its refusal wording, on `graph_client/` for the transport and the walk, and on nothing else.

Vocabulary is what two tools must not be free to disagree about — the `teams:///` grammar, the shape
of a message, who the signed-in user is. A window size and a paragraph are neither.

## Layering

No rule becomes newly assertable at this level. The one rule still absent (rule 7, no module
addresses a single meeting recording) needs a recordings listing to be the surface it protects, and
its anti-vacuity guard would still fail here; it arrives with the tool that lists them, exactly as
`tests/test_layering.py` says it will. Every rule that *is* asserted keeps working unchanged: this
tool imports `shared/seam.py`, `graph_client/` and FastMCP and nothing else of the package, names no
sibling, spells no handle, and enters every package through its `__init__`.

What does get exercised for the first time is rule 4's reason for existing rather than rule 4
itself. `list_teams` and `list_channels` are the first pair of tools where one's answer is the
other's argument, and they still share no code, no constant and no sentence — the connection between
them is a `team_id` on the wire and two paragraphs that each argue their own side of it.

## Fallout, deliberately included

`list_teams` shipped a description saying "No tool here takes a team id as an argument". It was true
when it was written and this PR makes it false, so it is corrected here, in its own commit. A
description is live protocol surface: a stale sentence does not send a model somewhere that fails
loudly, it talks the model out of a call that would have worked. No name, argument, field or
permission of `list_teams` changes — prose only.

### DIFF
--- a/services/office-mcp/README.md
+++ b/services/office-mcp/README.md
@@ -2,12 +2,12 @@
 
 An MCP server for Microsoft 365 via Microsoft Graph API.
 
-Users sign in with their own Microsoft account and the server acts as them. It exposes five MCP
+Users sign in with their own Microsoft account and the server acts as them. It exposes six MCP
 tools so far — `get_me`, the signed-in user's own profile; `list_chats`, their Microsoft Teams chats
-most recently active first; `list_teams`, the teams they are a member of; `search_messages`,
-full-text search across every Teams message they can see; and `read_message`, one of those messages
-in full — each one a file of its own, and more land in later PRs, stacked on top of this one, one
-tool per PR.
+most recently active first; `list_teams`, the teams they are a member of; `list_channels`, the
+channels of one of those teams; `search_messages`, full-text search across every Teams message they
+can see; and `read_message`, one of those messages in full — each one a file of its own, and more
+land in later PRs, stacked on top of this one, one tool per PR.
 
 ## Layout
 
@@ -78,6 +78,7 @@ Tools redeem permissions per call via On-Behalf-Of. No permission = no consent =
 | `User.Read` | Delegated | No | `get_me` |
 | `Chat.Read` | Delegated | No | `list_chats`, `search_messages`, `read_message` |
 | `Team.ReadBasic.All` | Delegated | No | `list_teams` |
+| `Channel.ReadBasic.All` | Delegated | No | `list_channels` |
 | `ChannelMessage.Read.All` | Delegated | Usually | `search_messages`, `read_message` (channels) |
 
 Multiple tools naming the same permission is normal. It is not duplication to remove. Each tool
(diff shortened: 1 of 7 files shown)


vabatta:

PR #799, opened 2026-08-15
feat(office-mcp): walk one channel’s posts and their newest replies

vabatta:
## What this answers

"What is going on in #release."

No tool on this server could answer it. `search_messages` finds messages by keyword across every
chat and channel at once and cannot be scoped to one of them; `read_message` reads a message
somebody already holds a handle for; `list_channels` names a channel and then stops. `browse_channel`
takes the `team_id` and `channel_id` that pair already returns and answers with the channel's posts,
each followed by the replies to it, every message whole — the same shape and the same
Teams-HTML-normalised text `read_message` returns, so nothing that comes back needs a second call to
read.

It adds no permission to the consent screen. `ChannelMessage.Read.All` is what `search_messages` and
`read_message` already declare, so `GRAPH_SCOPES` comes out of the registry unchanged: same tuple,
same order, same cached On-Behalf-Of token keys, and no tenant has to re-consent to deploy this.

## The decisions worth knowing

**One request, never a sweep.** Graph allows about one request a second on a given channel for the
whole app per tenant. That budget is per *app*, not per user, so a loop over a team's channels
degrades every other user of this connector, and following `@odata.nextLink` down one channel spends
that channel's whole tenant budget on one caller. So the tool issues exactly one request. `$top` is
the window; `$expand=replies` makes a thread part of that one request instead of a round trip per
post; both cursors Graph offers — the collection's, and a post's own `replies@odata.nextLink` — are
read and not followed. A caller who needs more raises `limit`, up to Graph's own ceiling. A caller
who needs to reach across channels was always meant to use `search_messages`, which covers every
channel for the price of the request it was making anyway. The description says all of this in as
many words, because a budget is only a bound if the caller can see it.

**The order is thread activity, not post recency.** Graph sorts this collection by the last modified
date of the *entire reply chain*, so a two-year-old post returns to the front the moment somebody
replies to it. The list is not reordered here — reordering would be inventing an order Graph did not
give — and the description tells the model to read `created_at` rather than the position, and not to
report the top of the list as the latest news in the channel. This is also the second reason there
is no walk: "stop paging once a page is older than X" is unsound on a collection ordered this way,
so paging is bounded by a count and never by a date.

**No date parameter, and that is Microsoft's limit rather than an omission.** `$top` and
`$expand=replies` are the only parameters this collection takes — no `$filter`, no `$orderby` — so
rather than faking a date bound out of client-side filtering, the tool has none and points at
`search_messages`, which puts the date into the search index instead. Two assertions in the tests are
about parameters that must *not* go on the wire, for exactly that reason.

**This is the one answer here that cannot say whether it was everything.** Everywhere else on this
connector a page short of `limit` is the end of the collection, because the walk underneath followed
Graph's paging to it. Here nothing is followed, and system messages are dropped out of the page after
Graph has counted them into it — so a short answer says neither "that was the channel" nor "there is
more". Graph's own `@odata.nextLink` on that page does say, accurately, so it is reported: read,
never followed.

It is reported as two fields rather than one, and only when asked for. Two, because
`more_posts_in_channel` ("the channel holds more, and nothing here reaches it — search back in
time") and `posts_cut_to_limit` ("this window closed over posts Graph did send — raise `limit`") have
opposite remedies, and one boolean over both is the ambiguous flag no answer on this server carries.
Opt-in, because a field that is null for almost every answer is one a model need not reason about;
only the caller whose question turns on completeness pays for it, and asking changes what is
reported and never what was read.

**Every claim the prose makes about Graph is settled by a test, not carried across.** One request
per browse whatever the channel holds; the cursor read and the second page never fetched; a post's
own reply cursor not chased; the reply-chain order preserved with `created_at` telling the truth;
the same page answering differently with and without `@odata.nextLink`; system messages dropped from
posts *and* from replies. The end-to-end walk goes teams → channels → posts → a reply read back by
its own handle, over the real MCP protocol, with each id taken from the previous answer.

## What it adds to `shared/`, and why that is vocabulary

**The channel-reply handle shape.** Graph addresses a reply in a channel thread *under* the post it
answers, and the search projection carries no `replyToId` — so a search hit that is really a reply
gets the root-post shape, which Graph answers 404 to. `browse_channel` walks a channel post by post
and therefore knows each reply's parent, which makes it the only tool on this connector that can mint
the shape at all. It still lives in `shared/handles.py`, with the other three, because a handle one
tool mints and another reads is only readable while there is exactly one definition of it: two
spellers would not look like a disagreement, they would look like a handle one tool produced and
another 404s. `read_message` gains the nested request that resolves it in the same PR, because a
shape whose reader does not exist is a dead link dressed as an answer.

**`MAX_REPLIES_PER_POST`.** One tool applies it; two others can only *describe* it. When a search hit
is a reply, `read_message`'s 404 has to tell the model where a reply's own handle can be obtained and
where that runs out, and `search_messages` has to say the same on the field that mints the failing
handle. A second spelling of the number would be a refusal sending a caller to a window the browser
does not actually reach — and nothing would fail. That is precisely the test for shared vocabulary:
two files would each need a copy, and a difference between the copies is a bug a caller can see.

Neither of these is a layer. `shared/` publishes no `__all__` and is entered by module, so every
consumer names at its import line which piece of vocabulary it depends on; the visibility is the
point. What stayed in the tool file is everything one tool could own — the name, the description, the
permission, the arguments, the answer shape, the Graph request and every refusal only this tool can
explain.

## Layering rules

**No rule becomes newly assertable here, and that is the honest answer.** The one rule of the
finished set still absent — nothing may address a single meeting recording — needs a recordings
listing to be the surface it protects, and arrives with the tool that lists them. The numbering is
left as it is so that arriving costs a class and changes nothing else.

What does change is how much rule 6 forbids. "One speller per handle family" now covers four shapes
minted between three tools, and its guard checks both halves: that `handles.py` still writes every
family the rule gives it, and that no other module under `src/` builds or parses one. The reply shape
is the strongest case the rule has — the only tool that can mint it and the only tool that resolves
it are different files, and neither spells the URI.

Rules 1 and 4 are the ones a tool like this is most likely to break, and both still hold by
construction: `browse_channel` imports `shared/`, `graph_client/` and FastMCP and nothing else of
this package, and the registry is still the one place every tool module is named — which is where the
consent screen's scopes come from.

### DIFF
--- a/services/office-mcp/README.md
+++ b/services/office-mcp/README.md
@@ -2,10 +2,13 @@
 
 An MCP server for Microsoft 365 via Microsoft Graph API.
 
-Users sign in with their own Microsoft account; the server acts as them. Six MCP tools are available:
-`get_me` (profile), `list_chats` (chats by recency), `list_teams` (teams the user is in), `list_channels`
-(channels in one team), `search_messages` (full-text search), and `read_message` (one message in full).
-Each tool is one file; more land in later PRs, one tool per PR.
+Users sign in with their own Microsoft account and the server acts as them. It exposes seven MCP
+tools so far — `get_me`, the signed-in user's own profile; `list_chats`, their Microsoft Teams chats
+most recently active first; `list_teams`, the teams they are a member of; `list_channels`, the
+channels of one of those teams; `browse_channel`, what was posted in one of those channels;
+`search_messages`, full-text search across every Teams message they can see; and `read_message`,
+one of those messages in full — each one a file of its own, and more land in later PRs, stacked on
+top of this one, one tool per PR.
 
 ## Layout
 
@@ -77,7 +80,7 @@ Tools redeem permissions per call via On-Behalf-Of. No permission = no consent =
 | `Chat.Read` | Delegated | No | `list_chats`, `search_messages`, `read_message` |
 | `Team.ReadBasic.All` | Delegated | No | `list_teams` |
 | `Channel.ReadBasic.All` | Delegated | No | `list_channels` |
-| `ChannelMessage.Read.All` | Delegated | Usually | `search_messages`, `read_message` (channels) |
+| `ChannelMessage.Read.All` | Delegated | Yes, in most tenants | `browse_channel`, `search_messages`, `read_message` (channels) |
 
 Multiple tools naming the same permission is normal. It is not duplication to remove. Each tool
 declares its own tuple because that tuple words its own 403 and AADSTS65001 messages.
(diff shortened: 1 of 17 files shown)


vabatta:

PR #800, opened 2026-08-15
feat(office-mcp): read a Teams meeting's transcript in speaker turns

vabatta:
Add the `read_transcript` tool. This tool turns a transcript handle into speaker-attributed, timestamped turns.

The tool reads the handle and does not resolve the meeting. The handle carries both ids needed for one Graph request to `/transcripts/{id}/content`. Resolving the meeting again would be a second request and a second permission.

The tool asks for `text/vtt` format first. If the tenant forbids speaker names, Graph answers with inner code `SpeakerAttributionNotAllowed`. The tool retries, asking for `application/vnd.microsoft.graph.transcript+text`. The tenant-wide switch that blocks all transcripts has no request-side workaround, so the retry is keyed on the inner code only.

Filters narrow the answer. The endpoint is one stream with no range contract. Every page refetches the whole transcript. Filters are applied to parsed turns before paging. The `next_offset` counts matching turns, not all turns.

The tool refuses an inverted time window before any request. The tool refuses a blank speaker before any request. Both would return an empty page that looks like a silent meeting.

Updates `list_meeting_transcripts` documentation. Names the new tool where the lister could not before. Records the shared permission.

## Role in the stack

This PR adds the reader for transcript handles. It builds on `office-mcp/feat/transcripts`, which lists transcripts for a meeting.

`read_transcript` completes the meeting path. A model calls `list_chats` to find meetings, `list_meeting_transcripts` to find their transcripts, and `read_transcript` to get the words — speaker-attributed and timestamped.

This is the surface where a delegated connector beats a link. No further work is needed on this tool.

### DIFF
--- a/services/office-mcp/README.md
+++ b/services/office-mcp/README.md
@@ -2,14 +2,15 @@
 
 An MCP server for Microsoft 365 via Microsoft Graph API.
 
-Users sign in with their own Microsoft account and the server acts as them. It exposes eight MCP
+Users sign in with their own Microsoft account and the server acts as them. It exposes nine MCP
 tools so far — `get_me`, the signed-in user's own profile; `list_chats`, their Microsoft Teams chats
 most recently active first; `list_teams`, the teams they are a member of; `list_channels`, the
 channels of one of those teams; `browse_channel`, what was posted in one of those channels;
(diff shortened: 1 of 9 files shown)


vabatta:

